Issue #3, opened 2020-01-10
Add device event emitters for WinRT, if possible


qdot:
WinRT currently only emits DeviceDiscovered. DeviceUpdated should be easy. However, DeviceConnected/Disconnected both happen in the context of the device itself, which can't see the Central object. We either need to attach event handlers to devices, or figure out some way to route that up through the Central handler for now.

qdot:
Fixed in January.

PR #6, opened 2020-01-10
rumble winrtble fixes

### LINKED ISSUE #5: WinRT should only emit discovery events after updating its properties
WinRT will emit advertisements before its updated the peripheral properties that it holds, meaning that we can have peripherals with no name or other identifying attributes.

blackspherefollower:
Fixes #5 and #3 
As well as the advertisement issue on the previous repo.


qdot:
ff98b7b was landed, I'd already bought in ffb1d8b as part of the merge fix.

PR #18, opened 2020-01-21
Allow parsing UUID and BDAddr from str

icewind1991:
Makes it easier to create a BDAddr/UUID by allowing parsing them from the same format as they are displayed


qdot:
Looks good, thanks!

I'm contemplating changing the B128 version to use the uuid module in the future, since that's already standardized, but this'll at least make things a bit easier to deal with until then.

PR #22, opened 2020-02-07
set dev_id when creating bluez adapter from device info

icewind1991:
PS: I noticed that my previous PR was set to the `dev` branch and not `master`, not sure if that was correct, I've put this one against `dev` too but things might need to be merged to master


qdot:
Dev was correct. :)

PR #23, opened 2020-02-07
pass LEAdvertisingData to event handlers

icewind1991:
For my use case the advertising data contains the broadcasted sensor data I'm interested in

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -21,6 +21,7 @@ use crate::{
     Result,
     api::UUID::{B16, B128}
 };
+use crate::bluez::protocol::hci::LEAdvertisingData;
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum AddressType {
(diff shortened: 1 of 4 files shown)


qdot:
This'll break all of the other platforms due to changing the format of the event enum, so I can't accept it until it's on all platforms. It'll hang out here until either someone does it or I have time to do it.

blandger:
Half of year has passes. Is it still relevant issue ?

qdot:
Yes, it's still relevant (please don't PM my issues, I'm working on cleanup right now). However, LEAdvertisingData only exists in/is emitted from bluez, so this data would need to be abstracted into a new cross-platform struct for this to work. Still figuring that out.

NZSmartie:
Since the HCI raw socket implementation has been removed and replaced with a BlueZ implementation, this PR is no longer relevant. Also with the recent #126, additional advertising data is now emitted as `CentralEvent`s. 

If you need more advertising data to be available in the btleplug API, please open an issue

PR #24, opened 2020-02-18
Added adpater discovery and ble peripheral discovery example

blandger:
Hi
I've added one of my research code. It can 
- discover local BLE adapter
- discover found peripheral device(s)
- connect it 
- dump characteristics
- disconnect from BLE

Code is tested on Linux as primary OS.


qdot:
Thanks!

Issue #28, opened 2020-03-12
Impl FnMut for NotificationHandler to Enable Stateful Behaviors


IamfromSpace:
**Feature Description**

Currently, NotificationHandler implement `Fn`, which allows user to pass in a closure.  This closure may have captured it's environment via `move`, however, this is a read-only capture.  If NotificationHandler was `FnMut`, it could modify variables it captured as well.

For example, this becomes possible:

```rust
let mut i = 1;
my_peripheral.on_notification(Box::new(move |n| {
  println!("This is notification number {:?}: {:?}", i, n);
  i += 1;
}
```

This can be especially useful in places where characteristics require two points in time to give meaningful data (ex. cycling cadence).  So the previous can be constantly captured and updated, to be compared against the new incoming notification.

I've got a pull request incoming for this :)

qdot:
Closing, as this was merged in March. But notifications will follow event handlers to being channels soon.

PR #29, opened 2020-03-12
Allow NotificationHandlers to be FnMut, so they can mutate anything m…

IamfromSpace:
…oved in. Solves #28.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -84,7 +84,7 @@ pub type Callback<T> = Box<dyn Fn(Result<T>) + Send>;
 pub type CommandCallback = Callback<()>;
 pub type RequestCallback = Callback<Vec<u8>>;
 
-pub type NotificationHandler = Box<dyn Fn(ValueNotification) + Send>;
+pub type NotificationHandler = Box<dyn FnMut(ValueNotification) + Send>;
 
 /// A Bluetooth UUID. These can either be 2 bytes or 16 bytes long. UUIDs uniquely identify various
 /// objects in the Bluetooth universe.
--- a/src/bluez/adapter/acl_stream.rs
+++ b/src/bluez/adapter/acl_stream.rs
@@ -22,6 +22,7 @@ use crate::{
         protocol::att,
         adapter::Adapter,
     },
+    common::util,
     Error,
     Result,
     api::{CommandCallback, RequestCallback, BDAddr, NotificationHandler, Characteristic},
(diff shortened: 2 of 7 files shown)


qdot:
Solves #28 for a single platform. :)

I think I'm ok with the idea of this overall, but this implementation should probably be promoted to a utility function that we can pass the handler Mutex to or something, so I can handle it in Mac/Windows too.

If this is going to be integrated, we're going to have to do this for all platforms, not just linux, so I can't land this until that happens.

IamfromSpace:
Ah, good point!  I got excited that it worked and pushed it :)

I'll take a look at solving this for other platforms too this evening.  And a utility makes sense as well, considering there's some complexity to it.

qdot:
Yeah no prob, I just use this as the cross platform layer of https://github.com/buttplugio/buttplug-rs so I'm very attuned to any breakage outside of Linux. :) 

Luckily, this *should* be easy to abstract up to the "API" level (terminology borrowed from rumble), and I believe all of the notification handlers across the platform look the same, so you shouldn't need to worry about hitting anything platform specific on Win/Mac I think.

I'm hoping to get unburied from a ton of stuff in the next week or so, at which point I'll be starting work on a redesign of this library with a more cross-platform focus. The current version was very much just a proof of concept and it's locked to bluez structures that it shouldn't be. :)

IamfromSpace:
I think this should now work on all platforms--however, it's seemingly very difficult for me to check on my linux device.  I hacked around and was mostly able to validate mac, but totally failed for windows.  Is there an way for me to validate checking on other platforms?

qdot:
I'll give it a shot on windows and let you know.

blandger:
Seems like all dev has stopped? No changes, no PRs?

IamfromSpace:
I’m fairly certain this code works—I just don’t have a means of confirming Windows behavior (or I’ll need to get more creative, haha).

Still open to any feedback/changes in the code, of course, but there aren’t any open concerns.

My focus shifted away from BLE for a bit, but I have some other issues that I patched in a fork that I’d eventually like to get upstream too.

qdot:
I'm alive, I'm just busy at the layer above this for me. 

I'm hoping to get back to this fairly soon, the problem is that the next step is basically "start writing documentation in order to burn this down and start again" because right now the project is definitely not sustainable (lots of crossed idioms, Arc<Mutex<T>>'d everything, etc), so it's kind of a lot to think about at the moment.

I'll give this a shot on windows tonight and try to get this plus the example PR in at least.

Issue #27, opened 2020-03-08
Derive Serde Serialize/Deserialize for All Types


IamfromSpace:
**Feature Description**

I'm working on an app right now that both displays notifications live and then records them for later processing.  Since I'm recording multiple characteristics, I need serialize-able tags to capture along with the notification value, and what better way than the `UUID` itself.

The `UUID` is a fairly trivial enum, so just creating an `SUUID` with `From` and `Serialize` implementations in my application solves my immediate needs.  However, it would simplify things and potentially be useful to others if this (and all or most other types) were Serializable as well.

I'm happy to contribute this feature (since it's more of a chore than anything else), if it seems like something worth adding!

qdot:
Came in with 0.4.2

PR #30, opened 2020-03-13
Derive serde Serialize and Deserialize…

IamfromSpace:
… for all public non-stateful structs and enums.  Solves #27.

Notably this isn't _all_ types, but this seemed to be a good balance.  Generally this cross-platform, public structs and enums that implement Copy.  It didn't seem so useful to allow serialization of things that are stateful, and the only other public types are within bluez and seem quite internal (maybe should be non-public or crate scoped).

There is some hit to compile times, it's 10% more for debug, and almost 25% more for release.  Not sure that should prevent this, but something worth considering.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ failure_derive = "0.1.1"
 backtrace = "0.3.5"
 async-std = { version = "1.4.0", features = ["unstable"] }
 uuid = "0.8.1"
+serde = { version = "1.0", features = ["derive"] }
 
 [dependencies.nom]
 version = "^4.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ failure_derive = "0.1.1"
 backtrace = "0.3.5"
 async-std = { version = "1.4.0", features = ["unstable"] }
 uuid = "0.8.1"
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0", features = ["derive"], default-features = false, optional = true }
 
 [dependencies.nom]
 version = "^4.0"
--- a/README.md
+++ b/README.md
@@ -164,6 +164,17 @@ Beyond that, some of our other goals are:
 | Write Descriptor (Sync) ||||
 | Write Descriptor (Async) ||||
 
+## Library Features
+
+#### Serialization/Deserialization
+
+To enable implementation of serde's `Serialize` and `Deserialize` across some common types in the `api` module, use the `serde` feature.
+
+```toml
+[dependencies]
+btleplug = { version = "0.4", features = ["serde"] }
+```
+
 ## Old rumble README Content
 
 ### Rumble
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -15,13 +15,15 @@ use std::{
     fmt::{self, Display, Formatter, Debug},
     collections::BTreeSet,
 };
+#[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 use crate::{
     Result,
     api::UUID::{B16, B128}
 };
 
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub enum AddressType {
     Random,
     Public,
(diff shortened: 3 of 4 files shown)


qdot:
Ok, so, a couple of things:

- Outside of your personal project, how do you see this as being useful to the library?
- Even if this is useful to the library, it should be behind a feature flag that's defaulted off.

IamfromSpace:
Yeah, I think it’s a good question, and hoping the PR process would vet out.

To start with, there are absolutely workarounds, and that’s what I have implemented at the moment.  You simply make a new type that mirrors it, and impl From back and forth.  As I have it now, it’s quite hidden.

However, other std types like Duration just serialize right out of the box and it’s really nice.  No random layer of indirection or boilerplate.

As to will others use it, I’m immediately biased because my sampling is 1/1, haha.  So I’m honestly not sure.

The added cost also seems fairly low (5-20s on my machine, which is no monster) from a clean build (and cargo caching is quite good).  But it’s certainly not 0.

The thing that pulls me towards this being a good idea is that BLE is an inherently serialized thing.  There’s so many different types and custom types, that a library that fully abstracted away serialization would be monumental and quite difficult to maintain.  So there’s some awareness that working with this lib means working with raw bytes.  So from that vein, easing serialization in other places makes some sense to me.

I’m not sure serialization is usually an opt in feature.  My only concern about the feature flag is that it’s perhaps surprisal.

No feelings hurt too if this PR is rejected—it’s about as low effort as it gets and I have workarounds for my needs.  I’m much more interested in the question you posed: is this the right feature to add?

onelson:
I think there's a good precedent for this, but I might tweak the way it's implemented a bit such that the derives are gated by a feature. Serde is ubiquitous, and since it's impossible to implement/derive Serialize/Deserialize for types that belong to _someone else's crate_, this would have to happen _here_ in this one.

Crates like `chrono` allow you to opt-in by enabling the `serde` feature (off by default).
I've done this myself when defining a crate for types that will be shared between an http server and a wasm client:

Should look something like this to conditionally add the derive attribute when the feature is enabled.

```
#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
```

qdot:
Yup! I use the same type of feature gated inclusion over in buttplug-rs, because we don't *always* need serde. 

I think with that addition, I'd probably be ok with this for now. I was asking about usage just to make sure I wasn't missing something that I should add to the doc I am about to mention. :3

BTW, just to give you an idea of my future plans:

The way the library stands now, it's... a mess. I just hacked 2 other platforms into Rumble (which was built specifically for Bluez and it shows) to get something working for cross-platform support in buttplug-rs in January, then got super busy. The super busy is wrapping up, so my plan is to finish my next version of buttplug-rs, then start a big ol' design and architecture document here to think about how things should work, between rust idioms and OS specific bluetooth API idioms. I've worked on cross platform bluetooth in C# and Node already, so I'm sadly way too familiar with this stuff. :(

To give some examples of what I'm thinking, the "notification handler" idiom seems to not fit the langauge. This should just be a channel we can fling events through, and expect the user to take care of events like a stream (somewhat similar to https://github.com/najamelan/pharos/ but I don't think this needs its own module). That saves us from having to deal with ownership like we do now. Similarly, for UUID, there's already UUID libraries out there that handle this sort of thing (though BLE's 2 byte UUID shortcut is annoyingly specific :| ), so I've been considering reusing another library (if needed, depending on what falls out of aforementioned architecture doc) instead of just bringing this UUID over.

While that's in flight, I'm still happy to bolt stuff on to the current version of btleplug as needed (I've gotta get device disconnect events working ASAP :( ), but it'd be nice to have something that feels a bit more planned than what I've got right now. Will definitely take input too. :)

IamfromSpace:
Yeah, I'm sold on the feature flag too with that example as a reference, I think that works nicely.  Also added a snippet to the README to make sure it was obvious (wasn't _exactly_ certain the best spot, so I just tried to make it work).

Makes a lot of sense on the current and future state!  I think I'd personally like to the see the interface more "BLE-ish" as well.  Little things like `request` and `command` vs `write` and `write_without_response` just add indirection.  I think it's not really this crate's job to abstract away the BLE elements, it should focus on providing a seamless GATT interface experience.  Then higher levels can offer more abstraction for _specific_ peripherals like `read_heart_rate` or `calibrate` or whatever.

From the GATT side of things, it would also be interesting to explore what the embedded folks have been doing with `embedded-hal` where there's common interfaces that are implemented across different hardwares.  Then theoretically if you wanted to use any device/OS as a central, as long as someone implemented the interface, you could!

Excited to see where this crate goes in the future :)

Issue #25, opened 2020-03-05
Cannot reconnect to Peripheral after Disconnect


IamfromSpace:
My code is setup to connect to a peripheral, subscribe to it, output its notifications to the console.  It then attaches a subscription to the central, which logs all events and attempts to reconnect to any address that appears in a DeviceDisconnected event.

I then execute the program, wait for notifications, physically shield the peripheral wait for the disconnection, then immediately unshield the device.

**Expected behavior**
The device should reconnect.

**Actual behavior**
There is a 20s timeout error.

**Additional context**
I have also tried adding 30s delays, trying to disconnect from the device before connecting, and trying to start and then stop scanning before re-connecting.  None of these things seem to work.  This is a Polar H10 HRM, I have a few other BLE devices I will also try.

qdot:
What OS are you on?

IamfromSpace:
Good question!  Sorry I left that out.  This is on Linux.

Also I tested this on a Kickr Core bike trainer, which can just be unplugged and plugged back in.  The effect is the same.

However, this has a blue LED with three states: blinking is not connected, fast blink is connecting, and solid is connected.  What’s interesting is that on the call to connect, the light goes solid!  So it appears that the device believes it’s connected, but the central does not.

IamfromSpace:
Also just tried this out on my Raspberry Pi Zero W and got the exact same result.

So this is replicated on:
Rasbian 2020-02-05-buster-lite + bluez 5.50
Ubuntu 16.04 + bluez 5.41

IamfromSpace:
So digging into this, it appears that the issue appears to be _half_ my fault, but this is very un-intuitive, and likely needs a fix.  The issue is that when I was listening to disconnects, I was immediately trying to connect back to that same device that was disconnected in a blocking fashion.  What's not so obvious is that:
  a) The event callback is on the same thread as the socket reading thread.
  b) A connection cannot succeed until it receives a connection success event from the socket.

Essentially, I caused the deadlock with my, "if you see a disconnect, reconnect" behavior.  I can theoretically fix this by handling this in a non-blocking way.

However, rather than put this requirement on a user, it likely makes more sense to call event handlers in a separate thread, so that no action by the user blocks the socket.

qdot:
As of v0.5.0, events are now sent on a channel, so this shouldn't be a problem anymore. Closing.

PR #32, opened 2020-03-15
Don't block the thread that's reading from the socket…

IamfromSpace:
… with the user's callbacks on events.  Solves #25.

I'm not sure that this a _great_ solve.  @qdot, this sort of gets to your point of maybe registering sets of callbacks just isn't the right approach as a whole for rust.  If this were a stream, you wouldn't really face this issue in the first place.

This isn't strictly necessary, as the user can basically do this exact same thing around their own handlers--it's just not obvious when one might want to do so.

### DIFF
--- a/src/bluez/adapter/mod.rs
+++ b/src/bluez/adapter/mod.rs
@@ -301,10 +301,12 @@ impl ConnectedAdapter {
     fn emit(&self, event: CentralEvent) {
         debug!("emitted {:?}", event);
         let handlers = self.event_handlers.clone();
-        let vec = handlers.lock().unwrap();
-        for handler in (*vec).iter() {
-            handler(event.clone());
-        }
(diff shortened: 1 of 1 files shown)


qdot:
Ok so the problem here is going to be differences in expectations for handlers and their thread contexts between platforms, since this only updates the linux handler. Holding off on this until I rip out the event system and replace it with something sane, which will hopefully be soon.

blandger:
Half of year has passes. Is it still relevant issue ?

qdot:
@blandger Yes, this is still relevant, and it's being fixed in v0.5.

qdot:
Callback event handlers are no longer a thing, replaced by channels so it's up to the user what context they want to handle in. Closing as WONTFIX, but thanks for the patch!

PR #33, opened 2020-03-15
Updated windows BLE Manager + examples

blandger:
- BL Manager for windows now returns list of adapters to be consistent with other version
- updated /examples for new signature, updated discover_adapters_peripherals.rs to be compiled in Linux + Windows OS
- minor warning fixes 

I can't really check windows code in run-time because I have only Windows 8 available, sorry

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 /target/
 **/*.rs.bk
 Cargo.lock
+**/.vscode*
+/workspace.xml
+**/*.code-workspace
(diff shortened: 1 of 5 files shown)


blandger:
DO NOT MERGE yet pls...

blandger:
Not example is ready.

qdot:
Thanks for the new examples! 

I had to hand merge this 'cause bringing in some other PR's before this caused conflicts, but it's on dev now.

This patch still leaves MacOS returning Result<Adapter> while Windows/Linux return Result<Vec<Adapter>> now, but I'll fix that up in a followup (and start enforcing it via traits soon hopefully).

Issue #16, opened 2020-01-12
Set up Azure Pipelines CI


qdot:
It's where I run all of my builds currently, works well for cross platform, etc.

mfiumara:
Is there a reason to use zure pipelines over github actions? Since the code is hosted on github might make sense.

qdot:
@mfiumara Mostly that I already have deep experience with Azure Pipelines and should be able to bring over my already built configs.

mfiumara:
Alright makes sense, I did some azure pipelines myself and it's a very nice CI tool. If you want to check how github actions look like I linked a PR.

qdot:
Thanks for the PR, but went with Azure anyways.

PR #37, opened 2020-03-27
Template github actions

mfiumara:
Template github action for CI. Related to #16 .


qdot:
Thanks, but pulling Repo Maintainer rank and going with azure pipelines. I have multiple packages on there that use this already, so it's easiest for me to track everything there.

PR #38, opened 2020-03-29
Correct comment regarding BlueZ sockets

wezm:
Small fix for comment that confused me for a second. The updated text is consistent with the rumble README quoted in this repo's README:

https://github.com/deviceplug/btleplug/blob/984e6cf363176927c008d8a5d26b01edac4517cf/README.md#rumble


qdot:
I can def see how that would be confusing. :D

PR #39, opened 2020-04-20
Correctly get peripheral local_name on macOS 10.15

windoze:
MacOS 10.15 (10.14?) somehow changed the process of peripheral discovering, we may not have the name when the first time a device is discovered, and the name comes with following `centralManager:didDiscover...` events.


qdot:
LGTM, thanks! Merged to dev. I'll try to get this released soon (as it'll also affect some of the stuff I'm working on.

Issue #40, opened 2020-04-21
Add an example for writing own Central::on_event handler


vietlq:
**Add an example for writing own Central::on_event handler**

Hi, I'm new to Rust and don't quite understand how to write an instance of `EventHandler` that can be passed to Central::on_event. Any quick example would be appreciated. I can update README.md via PR once you show me a basic example by replying here. Thank you.

```
pub type EventHandler = Box<dyn Fn(CentralEvent) + Send>;

/// Central is the "client" of BLE. It's able to scan for and establish connections to peripherals.
pub trait Central<P: Peripheral>: Send + Sync + Clone {
    /// Registers a function that will receive notifications when events occur for this Central
    /// module. See [`Event`](enum.CentralEvent.html) for the full set of events. Note that the
    /// handler will be called in a common thread, so it should not block.
    fn on_event(&self, handler: EventHandler);
```

Moret84:
Hi, 

If it helps, I have implemented scanning with callback in one of my project: [rs-mijiabt](https://github.com/Moret84/rs-mijiabt/blob/master/src/btleplug_ble_repo.rs) 

(I am quite new to rust too)

qdot:
The way the library currently handles events is not really what I'd call idiomatic rust (I didn't write it, it was adapted from one of the 3 libraries this library pulled from just so we could get cross platform bluetooth ASAP). Passing in closure callbacks is a nightmare for lifetime handling. This should use a stream that you can manage whenever you want, and that's the plan for the near-to-medium future.

For my library right now, I convert to a streaming method by using channels.

https://github.com/buttplugio/buttplug-rs/blob/master/buttplug/src/server/comm_managers/btleplug/btleplug_internal.rs#L62

Basically I just create a channel, and throw the sender into the event handler. Then I can deal with the events whenever I want by either waiting on/racing the receiver.

qdot:
Oh yeah, and leaving this open, as I'll see about adding an example for showing how that works so you don't have to go digging through my very odd libraries. :)

vietlq:
Thank you @Moret84 and @qdot for your kind responses.

qdot:
This landed in April, closing.

PR #42, opened 2020-04-23
Added `on_event` example

vietlq:
Added `on_event` example. Related to https://github.com/deviceplug/btleplug/issues/40

My VSCode also removed some trailing spaces, I hope you don't mind.

Please squash upon merging. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ pub fn main() {
 }
 ```
 
-Above use is just waits for 2 seconds to see whatever device it can discover. This code example shows how to use event-driven discovery:
+Above code use is just waits for 2 seconds to see whatever device it can discover. This code example shows how to use event-driven discovery:
 
 ```rust
 let (event_sender, event_receiver) = channel(256);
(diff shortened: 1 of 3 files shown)


vietlq:
@qdot do you want me to make more changes?

qdot:
I'm way behind on PRs right now, I'll try to look this over soon.

blandger:
Where the code  ` = channel(256);` comes from ? What is library is used for that function?

qdot:
That's using async_std channels (and tasks, which is a bad idea since it locks us to a runtime), which came in as part of the corebluetooth changes I made. I need to swap it to use futures::channel at some point soon, I'm just used to async_std and it was the easiest thing to get the library out the door.

blandger:
@qdot Thanks, I figured that out.

But I still can't figure out how to connect to BLE and start receiving data in loop, I would like to have ability to reconnect after link drops and continue receiving data. There is luck of good examples. I'm thinking about possible implementation ways.

I would like create kind of server app. It should keep BL connection with medical device(s), store data locally, interact with user by some UI (web I suppose).
Currently I can't do even easy tasks without examples. So I'm trying to figure out how to create simple code for such stuff.
Thank you for efforts. I would like to help with unit tests. I will keep going to create something by myself.

qdot:
Apologies for taking so long to bring this in. See #45 about the channel() calls too, which will hopefully be removed in v0.5.0 (which at my current pace will happen sometime between next week and the heat death of the sun.)

PR #47, opened 2020-06-13
Bugfix/bluez handle indications

### LINKED ISSUE #46: Indications (ack'd notifications) are never Ack'd on Linux
**Describe the bug**
A BLE peripheral has two ways that it can broadcast data to the central, it may notify or indicate.  Notify is a best effort, and may be dropped at any point.  Indicate will retry until it receives an ack.  On Linux, this ack is never sent, so the peripheral will assume there is an issue.

**Expected behavior**
The central can subscribe to indications.

**Actual behavior**
(For my tested devices) the subscriptions never create events and are never ack'd, so the peripherals disconnect after 30s due to the error.

**Additional context**
Indicate and Notify are not considered separate on other platforms.  For OSX and Windows, the difference and ack is handled internally.  This bug should only extend to Bluez.

IamfromSpace:
Fixes #46

### DIFF
--- a/src/bluez/adapter/acl_stream.rs
+++ b/src/bluez/adapter/acl_stream.rs
@@ -25,7 +25,7 @@ use crate::{
     common::util,
     Error,
     Result,
-    api::{CommandCallback, RequestCallback, BDAddr, NotificationHandler, Characteristic},
+    api::{UUID, CommandCallback, RequestCallback, BDAddr, NotificationHandler, Characteristic},
 };
 
 use std::{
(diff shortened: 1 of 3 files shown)


qdot:
Thanks! I still haven't had time to become familiar with the bluez core so I'm going to have to trust you on this one, but it compiles so that's good enough right? :)

IamfromSpace:
Haha, I suppose I could have offered more context too :). This is pretty well vetted and works nicely with my cycling trainer.  I compared to some other (appropriately licensed) BLE libs as well to ensure implementation was correct.

To send intensity commands to the trainer there’s a bit of a process to it, and you must specifically subscribe to an indication before it will let you command other characteristics.  This is the only device I was able to test, but that process has worked fully since.

IamfromSpace:
Also, I had rebased this onto dev, so by pulling it into master, this brought in all dev commits; just want to call that out in case it was not intentional.

This has some interaction with the FnMut feature, so the fix before and after looks quite different.  Depending on how you’d like to go about it, I can try to back port a version for master only—I just noticed after and it has some trickier merge implications.

qdot:
Oh no! I totally wasn't looking at the base branch when I merged. It looks like your base branch is still live, so I'll just add your remote and get it fixed up.

IamfromSpace:
Yeah, sorry, since the last one went to dev I was trying to be helpful, looks like it wasn’t so much!

I’d recommend a hard reset back to the previous master—this fix doesn’t merge cleanly with it unfortunately!  I wouldn’t recommend trying to address the conflicts, it’ll be a headache.

Depending on how long you think the dev to master will take I can make up some commits that will merge better :)

qdot:
All fixed up, commit is now on dev, will merge to master on 0.4.4 (which, if I don't get any other PRs, may just happen this week 'cause indication handling is important)

But just to make sure we're on the same page here: for the main repo, master and dev, before this PR, were in sync. I hadn't added anything to 0.4.3, and your commit merged fine to master which is why this happened in the first place. Is your repo in a different state or something?

IamfromSpace:
Yeah, my repo might be a little odd—its primary purpose is the branch with specific stuff for my project.

PR #49, opened 2020-06-28
Use const rather a mut object to write

oflebbe:
I looked into the code in order trying to understand how to use Bluez. 
This part stuck as it looked weird. Why should we use a mutable object for libc::write ?
Monkey patched it to use a const object

### DIFF
--- a/src/bluez/adapter/mod.rs
+++ b/src/bluez/adapter/mod.rs
@@ -383,11 +383,11 @@ impl ConnectedAdapter {
         }
     }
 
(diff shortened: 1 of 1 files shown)


qdot:
Thanks for the PR! There's a TON of stuff that shouldn't be mut, but the library is currently waiting on me to finish other things before doing a refactor to clean stuff like that up.

I'd also recommend checking out Blurz if you're looking to learn bluez stuff.

PR #64, opened 2020-07-28
Remember data from previous advertisements

rvolgers:
Right now, as soon as one piece of data comes in, all previous ones
are forgotten. This makes it impossible to match devices by name, if
more data is received after the advertisement containing the name.

This fixes the problem on linux and windows. I've made no changes to the
mac code, as I'm not sure it even implements this.

This is lightly tested at the moment. I ran unit tests, and the linux
changes are copied by hand from a more comprehensively modified local
version, on which they work (and fix the described problem).

Windows changes are trivial but have not been tested at all.

I added a number of comments to help others trying to work with the
code. Some may be subjective, in which case, change them as desired.

### DIFF
--- a/src/api/adapter_manager.rs
+++ b/src/api/adapter_manager.rs
@@ -75,14 +75,21 @@ impl<PeripheralType> AdapterManager<PeripheralType> where PeripheralType: Periph
   }
 
   pub fn add_peripheral(&self, addr: BDAddr, peripheral: PeripheralType) {
-      if self.peripherals.contains_key(&addr) {
-          panic!("Adding a peripheral that's already in the map.");
-      }
+      assert!(!self.peripherals.contains_key(&addr), "Adding a peripheral that's already in the map.");
+      assert_eq!(peripheral.address(), addr, "Device has unexpected address."); // TODO remove addr argument
       self.peripherals.insert(addr, peripheral);
   }
 
-  pub fn update_peripheral(&self, _addr: BDAddr, _peripheral: PeripheralType) {
-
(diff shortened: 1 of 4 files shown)


qdot:
Oops, thanks! I meant to add advertising comparison/updates before releasing 0.5.0 then totally flaked. Will check this later today and see about getting it landed and a v0.5.1 out.

rvolgers:
Cool, thanks! This is kinda rough though (I noticed right after submitting I messed up one of the asserts, so it wouldn't actually work... not a good sign. I fixed that one, at least.)

I'll see if I can hook it up to my code that actually exercises this and give it a spin later, but just wanted to get this out there so I don't forget.

qdot:
Windows didn't compile, there was a reference to self in a move closure. Easy fix 'cause it was for something I'd already cloned in, so I fixed it. :)

The FIXMEs are understandable, and I'm not real happy with how the system works right now, this whole change was just me getting mad enough to consolidate some of it. I'd honestly like to hold advertisement data separately and not create a peripheral instance until it connect, to at least give a bit of an RAII feeling. The update call should also not build a new object, but rather just update fields like RSSI, which are only updated via advertisements.

That said, I'm trying to avoid snowballing a bunch of stuff into the current api, that at some point in the (god I hope it's) near future will be overhauled and set into cross-platform traits versus this incredibly weird per-platform object thing, hence all of these small fixes right now.

However, it seems people other than me have decided to use the library too, so we'll see how long I actually get to hold out on that. :)

qdot:
Patch landed to dev, closing manually just because this was pointed at master.

PR #79, opened 2020-09-11
feat: Add write with response and read on Core Bluetooth

olefriis:
As the title says. Plus, for debugging purposes, I had to get the message from the error objects that get passed by Core Bluetooth on failure. I thought this would be a nice addition to the code, so we get more context than just `error`.

Please note: I am a total Rust n00b, so there may be a lot of stuff in this PR that is not idiomatic Rust. Please bare with me. I am super interested in fixing that, though, so any comments are welcome.

I've tried to keep the current code style.

Oh, and one last thing: It would totally make sense to split this PR into two - one for adding read support, another for adding write with response. However, the BLE device I am trying to get working (a smart radiator valve) requires both to be in place before I know it's working, so I am unable to do the split and tell if either works by itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Beyond that, some of our other goals are:
 | GATT Server Disconnect Event |X|X|X|
 | Write to Characteristic (Sync) |X|X|X|
 | Write to Characteristic (Async) |||X|
-| Read from Characteristic (Sync) |X||X|
+| Read from Characteristic (Sync) |X|X|X|
 | Read from Characteristic (Async) |X|||
 | Subscribe to Characteristic (Sync) |X|X|X|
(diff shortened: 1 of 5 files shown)


qdot:
Thanks for the PR!

This looks good! There's a few println!'s I'll clean up in a followup commit, but I think I can just bring this in as is (because as of about 3 hours ago I now need read on macOS too, so fantastic timing :) ).

As for being new to rust and working on this library, I... wouldn't really recommend this as a shining example of how to write rust. I wasn't kidding in the readme when I said this thing is a frankenstine'd mess. I'm hoping to improve the structure at some point, but your patch looks a lot like my work on this right now, just trying to get things far enough along to work for the moment, which is completely fine. :)

PR #80, opened 2020-09-27
Simplify accepted inputs to UUID::from_str()

aklajnert:
This PR is to broaden accepted values for `UUID::from_str()`.  
Usually we encounter UUID like this: `2f029dc0-98ff-40e4-aa1e-628135afe1ed`. To make it acceptable by `UUID::from_str()` we'd have to convert it to `2F:02:9D:C0:98:FF:40:E4:AA:1E:62:81:35:AF:E1:ED`. This change will accept both inputs.  
The conversion back `to_string()` will still produce `2F:02:9D:C0:98:FF:40:E4:AA:1E:62:81:35:AF:E1:ED`.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -185,8 +185,14 @@ impl FromStr for UUID {
 
     fn from_str(s: &str) -> ParseUUIDResult<Self> {
         let bytes = s
-            .split(':')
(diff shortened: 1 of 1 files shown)


qdot:
Wooo, and tests! Thanks :D

Issue #68, opened 2020-07-29
Rework BDAddr + UUID structs representaion


blandger:
**BDAddr + UUID structs representaion**

I would like to rework BDAddr + UUID structs internal 'reversed' representation by removing that piece of code (in format/display/debug + parse parts). It confuses a lot. Also I would like to add parse UUID as string variant(s).

BDAddr might be stored as MAC:
https://www.novelbits.io/bluetooth-address-privacy-ble/

characteristic for UUID might be:
_A universally unique identifier (UUID) is a 128-bit (16 bytes) number_
https://learning.oreilly.com/library/view/getting-started-with/9781491900550/ch04.html

MattesWhite:
To clarify I don't wanted to be the one refactoring `UUID`s and `BDAddr`, I just offered my help if it is appreciated. It is just the case that I have similar problems in my use case. I didn't felt offended, everything is fine.

Now back to the issue.

---

As you already noted about format/display/debug and parse, I'd like to give my list of requirements/features/wishes:

- Implement not only `Display` but also other [`std::fmt` traits](https://doc.rust-lang.org/std/fmt/#traits) like `LowerHex` and `UpperHex`.
- The default display should be with delimiters, i.e. `BDAddr` with `:` and UUID with hyphens.
- Optional formatting functions for serialization without delimiters.
  - This is something I used quite often as I use `BDAddr` and `UUID`s in the routes of my HTTP-server.
- The same for parsing: The parsers should not rely on the presence of delimiters and be case insensitive for hex digits.

This is only a list of what I would have tried to do. I don't insist on them and I'm open to discussion. Of course I will respect the wishes/vision of the project owner :wink: .

blandger:
Good additions, Mattes. I don't mind you can take that issue. Let Kyle decide by himself.

I can take issue upgrading project to NOM 5.1 version. I tried that locally, but it needs more research, because difference between 4 and 5 quite significant.

qwandor:
@MattesWhite's refactor of BDAddr is now merged, is an improvement.

However, we could consider instead using one of the existing crates which provides a standard MAC address type, such as [eui48](https://crates.io/crates/eui48) or [macaddr](https://crates.io/crates/macaddr). I haven't had a chance yet to see which of the two is a better choice. Any opinions?

MattesWhite:
[...]

Issue #76, opened 2020-09-04
No RSSI


Beetix:
**Describe the bug**
Can't get RSSI

**Expected behavior**
RSSI information for each message received

**Actual behavior**
No RSSI information

**Additional context**
RSSI mentioned [here](https://github.com/deviceplug/btleplug/blob/ae82cd986b5b5e9c29d0525a0de7789d0b81c5e8/src/api/adapter_manager.rs#L117) to set tx_power_level. Don't quite understand why. tx_power_level should be retrieved from the message sent by the peripheral while the RSSI should be retrieved from the central's radio.

qdot:
What platform/OS is this on?

Beetix:
This is on Arch Linux x86_64.

Is the RSSI information really implemented? If so, what is the current way of retrieving it?

Thanks

MattesWhite:
Hi,
I need the RSSI information for a current project. If someone gives me a rough instruction what is necessary to implement this feature I can write a PR.

qdot:
We get RSSI in all major platforms as of v0.8, closing.

rib:
Can we maybe re-open this issue, since I think TX power was maybe being considered as equivalent to RSSI here? 

The current status looks like:
In the bluez backend `tx_power_level` is correctly reported
In the winrt backend `tx_power_level` is actually the RSSI
In the corebluetooth backend `tx_power_level` is always `None`

It looks like `DeviceInfo` in the bluez backend includes the rssi which could be easy to export
In the winrt backend there is rss and tx power state under `BluetoothLEAdvertisementReceivedEventArgs` so `tx_power_level` can be updated to reflect the transmission power and it could also export the receiver power.
I'm not familiar with how to get either the tx or rssi state in corebluetooth, but _guess_ it's probably easily accessible somehow.

rib:
looking at the osx api I see that the tx is exposed via:
https://developer.apple.com/documentation/corebluetooth/cbadvertisementdatatxpowerlevelkey

The rssi is exposed asynchronously via:
https://developer.apple.com/documentation/corebluetooth/cbperipheral/1519111-readrssi

(and then it depends whether on ios or osx exactly how the rssi value is received)

qwandor:
Now fixed in dev by #198 for Windows and Linux, CoreBluetooth is still pending.

Issue #48, opened 2020-06-25
Stop using failure as a library error type


qdot:
**Describe the bug**
Switch from using failure to using thiserror, so the library doesn't leak types.

blandger:
I've tried github.com/shepmaster/snafu . As I see it's quite similar to thiserror. That doesn't matter for me which one can be used. I'm just curious about 'selection criterias'.

qdot:
I use thiserror over in buttplug-rs. It's a thinner layer on top of errors than snafu, just kinda felt better to me (not to mention I use color-eyre at the binary level which snaps in nicely). Snafu is also a fine library though, shepmaster does good work. Mostly a personal pref thing.

qdot:
Fixed in #108

PR #81, opened 2020-09-28
Refactor BDAddr

MattesWhite:
This PR follows the discussion of #68.

At last I found some time to work on this. For now I refactored `BDAddr` as a show case what I intend to do. If the implementation is approved I'll start working on `UUID`. Of course I'll include the changed suggested in #80 as well.

Initially, I wanted to write the parsers with `nom` but I recognized that there is a bunch of `nom v4` code in the crate so I settled for a 'hand-crafted' version. Another solution would be to use `regex` but I didn't wanted to pull in another dependency without asking.

### DIFF
--- a/src/api/bdaddr.rs
+++ b/src/api/bdaddr.rs
@@ -148,7 +148,7 @@ impl BDAddr {
     pub fn from_str_delim(s: &str) -> Result<Self, ParseBDAddrError> {
         let bytes = s
             .split(':')
-            .map(|part: &str| u8::from_str_radix(part, 16).map_err(ParseBDAddrError::InvalidDigit))
+            .map(|part: &str| u8::from_str_radix(part, 16))
             .collect::<Result<Vec<u8>, _>>()?;
 
         if bytes.len() == 6 {
--- a/src/api/bdaddr.rs
+++ b/src/api/bdaddr.rs
@@ -136,11 +136,13 @@ impl BDAddr {
     pub fn into_inner(self) -> [u8; 6] {
         self.address
     }
+
     /// Check if this address is a randomly generated.
     pub fn is_random_static(&self) -> bool {
         self.address[5] & 0b11 == 0b11
     }
-    /// Parses a Bluetooth address colons `:` as delimiters.
+
+    /// Parses a Bluetooth address with colons `:` as delimiters.
     ///
     /// All hex-digits `[0-9a-fA-F]` are allowed.
     pub fn from_str_delim(s: &str) -> Result<Self, ParseBDAddrError> {
@@ -157,6 +159,7 @@ impl BDAddr {
             Err(ParseBDAddrError::IncorrectByteCount)
         }
     }
+
     /// Parses a Bluetooth address without delimiters.
     ///
     /// All hex-digits `[0-9a-fA-F]` are allowed.
@@ -174,6 +177,7 @@ impl BDAddr {
         }
         Ok(Self { address })
     }
+
     /// Writes the address without delimiters.
     pub fn write_no_delim(&self, f: &mut impl fmt::Write) -> fmt::Result {
         for b in &self.address {
@@ -181,6 +185,7 @@ impl BDAddr {
         }
         Ok(())
     }
+
     /// Create a `String` with the address with no delimiters.
     ///
     /// For the more common presentation with colons use the `to_string()`
--- a/src/api/bdaddr.rs
+++ b/src/api/bdaddr.rs
@@ -1,15 +1,12 @@
 //! Implementation of Bluetooth's MAC address.
 
-use std::fmt;
+use std::fmt::{self, Debug, Display, Formatter, LowerHex, UpperHex};
 use std::str::FromStr;
-use thiserror::Error;
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "serde")]
 use serde_cr as serde;
-
-use crate::Error;
 
 /// Stores the 6 byte address used to identify Bluetooth devices.
 #[cfg_attr(
@@ -18,27 +15,26 @@ use crate::Error;
     serde(crate = "serde_cr")
 )]
 #[derive(Copy, Clone, Hash, Eq, PartialEq, Default)]
-#[repr(C)]
 pub struct BDAddr {
-    address: [u8; 6usize],
+    address: [u8; 6],
 }
 
-#[derive(Debug, Error, Clone, PartialEq)]
+#[derive(Debug, thiserror::Error, Clone, PartialEq)]
 pub enum ParseBDAddrError {
     #[error("Bluetooth address has to be 6 bytes long")]
     IncorrectByteCount,
-    #[error("All digits in a Bluetooth address must be hex-digits [0-9a-fA-F]")]
-    InvalidDigit,
+    #[error("Invalid digit in address: {0}")]
+    InvalidDigit(#[from] std::num::ParseIntError),
 }
 
-impl fmt::Display for BDAddr {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        <Self as fmt::LowerHex>::fmt(self, f)
+impl Display for BDAddr {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        <Self as UpperHex>::fmt(self, f)
     }
 }
 
-impl fmt::LowerHex for BDAddr {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl LowerHex for BDAddr {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         let a = &self.address;
         write!(
             f,
@@ -48,8 +44,8 @@ impl fmt::LowerHex for BDAddr {
     }
 }
 
-impl fmt::UpperHex for BDAddr {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl UpperHex for BDAddr {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         let a = &self.address;
         write!(
             f,
@@ -59,9 +55,9 @@ impl fmt::UpperHex for BDAddr {
     }
 }
 
-impl fmt::Debug for BDAddr {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        <Self as fmt::Display>::fmt(self, f)
+impl Debug for BDAddr {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        <Self as Display>::fmt(self, f)
     }
 }
 
@@ -119,12 +115,6 @@ impl From<BDAddr> for u64 {
     }
 }
 
-impl From<ParseBDAddrError> for Error {
-    fn from(e: ParseBDAddrError) -> Self {
-        Error::Other(format!("ParseBDAddrError: {}", e))
-    }
-}
-
 impl FromStr for BDAddr {
     type Err = ParseBDAddrError;
 
@@ -142,11 +132,9 @@ impl FromStr for BDAddr {
 }
 
 impl BDAddr {
+    /// Destruct the address into the underlying array.
     pub fn into_inner(self) -> [u8; 6] {
         self.address
-    }
-    pub fn bytes(&self) -> &[u8; 6] {
-        &self.address
     }
     /// Check if this address is a randomly generated.
     pub fn is_random_static(&self) -> bool {
@@ -158,9 +146,7 @@ impl BDAddr {
     pub fn from_str_delim(s: &str) -> Result<Self, ParseBDAddrError> {
         let bytes = s
             .split(':')
-            .map(|part: &str| {
-                u8::from_str_radix(part, 16).map_err(|_| ParseBDAddrError::InvalidDigit)
-            })
+            .map(|part: &str| u8::from_str_radix(part, 16).map_err(ParseBDAddrError::InvalidDigit))
             .collect::<Result<Vec<u8>, _>>()?;
 
         if bytes.len() == 6 {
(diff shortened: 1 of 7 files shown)


qdot:
I definitely like the direction this is going, though do note that your current version is failing on macOS and Windows.

I think you're fine to move on to UUID, though note that I did bring in #80 (as I'm about to do a quick point release to fix Windows UWP reads), but you can integrate that in as you need.

The `nom` dependency thing is killing me too, I'm really hoping to get back to this one day and possibly just move us to actual dbus usage. :(

I don't think bringing in `regex` would be a huge issue? But being an ex browser dev I'm also kind of oblivious to compile times and all that, and the library I use this for is... pretty violently monolithic so I have like 30+ dependencies there. :)

MattesWhite:
The `BDAddr` part is now finalized. All pipelines should only fail because of legacy code by now.

A question of style: Do you prefer having bigger modules (`src/api/mod.rs` has now more than 500 lines without tests) or smaller once, e.g. put `BDAddr` and `UUID` into submodules and re-export them? Personally, I'd prefer the latter but I'll align with your choice of style.

Once, I did a refactoring from `nom v4` to `nom v5` so I'm willing to take care of this. However, the issue of having no RSSI values available (#76) is way more pressing for me. In addition, there is already worked on [`nom v6`](Geal/nom#1132) which probably will improve support for custom errors so waiting a little bit and solving #48 first is maybe not that bad.

For now I'll focus on refactoring UUIDs.

qdot:
Looks like everything built on the latest commit, so that's positive.

And smaller modules, please! The big modules came in with the libraries I imported, I'm much more a fan of submodules and re-exporting.

For the nom stuff, it's really hard to say how much that matters other than the conflicting dependencies, especially if we start looking toward dbus again for bluez. The RSSI is now important for me too (My library that depends on this now advertises RSSI relay capabilities, and yet, I have no RSSI here... >.> ), so hopefully that can be looked at soon.

MattesWhite:
So first work on UUIDs is done. I decided that its most practical to use [`uuid::Uuid`](https://docs.rs/uuid/0.8.1/uuid/struct.Uuid.html) wherever possible but I kept the short-value variant which has nice benefits like use parse and format functions of `uuid` crate.

I can't guarantee that the Windows version still works or that I didn't introduce bugs. I tried to adapt what was possible but when I the need to reverse bytes here and there makes my head hurt. In addition, I have only a Linux machine to test, so I don't even know if it compiles. Is there a way to do (at least) a `cargo check` for another OS? All in all, it would be really nice if someone could verify the code on a Windows machine.

qdot:
After the issues I had with the macOS read/write patches, I'll be testing this throughly across all platforms before landing, because it directly affects users of my libraries that depend on this. 👍 

So far, according to CI it looks like the only thing catching you on windows is a &u16 to u32 miscast, so not *too* horrible?

MattesWhite:
Yes the fixing errors caught by the pipelines is easy but I don't wanted to fix each error separately, push a single commit, wait for CI, iterate and so on. This also costs pipeline time, I think? What really bothers me is that I often read that Windows GUIDs are little endian and the `uuid::Uuid` is big endian so I can't guarantee that the conversion functions do the right thing. And writing tests without IDE support is not very pleasant.

qdot:
Well, looks like you pass CI now at least. And I all do all my development on windows (and the vast number of users of this library are also on windows, and probably don't know they're using this library), so like I said, I'll at least give it a runthru before landing. :)

johnmave126:
Actually, how about using [eui48](https://crates.io/crates/eui48) for mac address? That crate seems pretty slim and seems to be treated as de-facto mac address crate in crate like tokio-postgres.

qwandor:
@MattesWhite Hi! Any chance you could rebase the BDAddr changes against the latest dev branch? The diff is a bit hard to follow at the moment.

I've done something a bit similar for UUIDs in #123, but using the type from the `uuid` crate directly in the API and internally, which makes equality work properly, and then using an extension trait and a helper function to convert to and from the short form where needed.

NZSmartie:
I've rebased it on to the current `dev` branch and omitted all UUID related changes. It did drop the useful UUID constants you had defined, but I'm sure we could add that in another PR, that way this PR only focuses on `BDAddr`

src/corebluetooth/adapter.rs
```diff
@@ -17,9 +17,8 @@ pub struct Adapter {
 }
 
 pub(crate) fn uuid_to_bdaddr(uuid: &String) -> BDAddr {
-    BDAddr {
-        address: uuid.as_bytes()[0..6].try_into().unwrap(),
-    }
+    let b: [u8; 6] = uuid.as_bytes()[0..6].try_into().unwrap();
+    BDAddr::try_from(b).unwrap()
```
NZSmartie, line 21: One of refactors that didn't quite survive the rebase was this. I may be worth reviewing as I ended up applying a dirty fix so that macOS/iOS would still build

src/winrtble/utils.rs
```diff
@@ -78,37 +62,8 @@ pub fn to_char_props(_: &GattCharacteristicProperties) -> CharPropFlags {
...
-    #[test]
-    fn check_address() {
-        let bluetooth_address = 252566450624623;
-        let addr = to_addr(bluetooth_address);
-        let result = to_address(addr);
-        assert_eq!(bluetooth_address, result);
-    }
-
-    #[test]
-    fn check_uuid_to_guid_conversion() {
```
qwandor, line 94: These UUID tests are still useful, please don't delete them.

NZSmartie: oh right, missed them when the UUID code got moved into its own module

MattesWhite:
Hi all,

sorry for the late response. Lately I don't have much time to monitor the progress of `btleplug`. However, I'm very exited to see the progress on `async` :+1: .

Feel free to edit this PR as much as you like. I'm happy that at least some pieces of my code end up in the crate. If there are any questions I'll be happy to help.

src/api/mod.rs
```diff
@@ -207,8 +142,8 @@ pub struct Characteristic {
     pub properties: CharPropFlags,
 }
 
-impl Display for Characteristic {
-    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+impl fmt::Display for Characteristic {
```
qwandor, line 145: What's wrong with importing `Display`. `Debug`, `Formatter` and so on as we were before?

src/api/mod.rs
```diff
@@ -385,28 +320,7 @@ pub trait Central: Send + Sync + Clone {
...
-            assert_eq!(result, expected);
-
-            if let Ok(uuid) = result {
-                assert_eq!(input, uuid.to_string());
-            }
-        }
-    }
+    // use super::*;
+
+    // No tests, yet.
```
qwandor, line 325: Just remove the tests module, rather than keeping it here empty.

src/lib.rs
```diff
@@ -108,4 +108,4 @@ pub enum Error {
 }
 
 // BtlePlug Result type
-pub type Result<T> = result::Result<T, Error>;
+pub type Result<T, E = Error> = result::Result<T, E>;
```
qwandor, line 111: This seems like a weird change to make. If you want a result with a different error type somewhere, can't you just use `std::result::Result` there?

src/api/bdaddr.rs
```diff
@@ -0,0 +1,270 @@
...
+use crate::Error;
+
+/// Stores the 6 byte address used to identify Bluetooth devices.
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(crate = "serde_cr")
+)]
+#[derive(Copy, Clone, Hash, Eq, PartialEq, Default)]
+#[repr(C)]
```
qwandor, line 21: Is this type being used for FFI somewhere?

src/api/bdaddr.rs
```diff
@@ -0,0 +1,270 @@
...
+pub enum ParseBDAddrError {
+    #[error("Bluetooth address has to be 6 bytes long")]
+    IncorrectByteCount,
+    #[error("All digits in a Bluetooth address must be hex-digits [0-9a-fA-F]")]
+    InvalidDigit,
+}
+
+impl fmt::Display for BDAddr {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        <Self as fmt::LowerHex>::fmt(self, f)
```
qwandor, line 36: Uppercase is more common, can we keep that as the default?

src/api/bdaddr.rs
```diff
@@ -0,0 +1,270 @@
...
+            Self::from_str_no_delim(s)
+        }
+    }
+}
+
+impl BDAddr {
+    pub fn into_inner(self) -> [u8; 6] {
+        self.address
+    }
+    pub fn bytes(&self) -> &[u8; 6] {
```
qwandor, line 148: Nit: add a blank line between each method.

qwandor, line 148: This method seems to be the same as `as_ref` above. No need for both.

src/api/bdaddr.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        let mut address = [0; 6];
+        for i in (0..12).step_by(2) {
+            let part = &s[i..i + 2];
+            address[i / 2] = u8::from_str_radix(part, 16).expect("Checked upfront");
+        }
+        Ok(Self { address })
+    }
+    /// Writes the address without delimiters.
+    pub fn write_flat(&self, f: &mut impl fmt::Write) -> fmt::Result {
```
qwandor, line 193: Is this actually used anywhere? I'd be inclined to leave it and `to_string_flat` out unless you have a usecase in mind.

src/api/bdaddr.rs
```diff
@@ -0,0 +1,270 @@
...
+        if s.len() != 12 {
+            return Err(ParseBDAddrError::IncorrectByteCount);
+        }
+        if s.bytes().any(|b| !b.is_ascii_hexdigit()) {
+            return Err(ParseBDAddrError::InvalidDigit);
+        }
+
+        let mut address = [0; 6];
+        for i in (0..12).step_by(2) {
+            let part = &s[i..i + 2];
```
qwandor, line 187: You can probably do this more nicely with `s.chunks(2)`.

src/api/bdaddr.rs
```diff
@@ -0,0 +1,270 @@
...
+            address[i / 2] = u8::from_str_radix(part, 16).expect("Checked upfront");
```
qwandor, line 188: Rather than checking twice, remove the `is_ascii_hexdigit` check above, and return the error here instead.

src/api/bdaddr.rs
```diff
@@ -0,0 +1,270 @@
...
+            cpy.copy_from_slice(&slice[..6]);
+            Ok(cpy.into())
+        }
+    }
+}
+
+impl From<u64> for BDAddr {
+    fn from(int: u64) -> Self {
+        let mut cpy = [0; 6];
+        let slice = int.to_be_bytes(); // Reverse order to have MSB on index 0
```
qwandor, line 108: Check that the top two bytes are 0.

qwandor:
I think that this is generally an improvement, once the a bunch of small things are fixed. However, we could consider instead using one of the existing crates which provides a standard MAC address type, such as [eui48](https://crates.io/crates/eui48) or [macaddr](https://crates.io/crates/macaddr). I haven't had a chance yet to see which of the two is a better choice.

src/api/mod.rs
MattesWhite: That's just personal preference, rather use the module name than importing everything. Especially, as we have to use `fmt::Result` anyway.

However, if this is contrary to the common style there is no problem in importing stuff from `std::fmt` directly.

src/lib.rs
MattesWhite: Again just personal preference, I have started a [thread in URLO](https://users.rust-lang.org/t/dont-do-type-result-t-std-result-t-myerror/49421) about that topic. tl;dr Why should I be required to use `std::result::Result` just to state that other type parameters are used.

... back when I wrote the PR I was not that aware of 'keep a PR small and dedicated to a problem and keep personal preference out' or however it is expressed best (I'm not an english native speaker).

src/api/bdaddr.rs
MattesWhite: That is legacy from rumble (?). I wouldn't mind removing it.

src/api/bdaddr.rs
MattesWhite: Or a comment would be even better :grin:

MattesWhite: Yeah, if I remember correctly, this was a legacy function but if we are rewriting the whole thing anyway, we can remove it as well.

src/api/bdaddr.rs
MattesWhite: Actually this was my initial motivation to write this PR (see https://github.com/deviceplug/btleplug/issues/68#issuecomment-666211155). I'm working on a 'Bluetooth gateway' where the BDAddr of devices is used as the identifier in a REST API. As the default `:` in MAC addresses is also a valid separator in URIs I use this _flat_ format. `eui48` provides a (similar format](https://docs.rs/eui48/1.1.0/eui48/struct.MacAddress.html#method.to_hexadecimal).

Don't know if this is enough to justify its existence in `btleplug`.

src/api/bdaddr.rs
MattesWhite: Is this really a problem/error? Would you rather use `TryFrom`?

src/api/bdaddr.rs
MattesWhite: At least I renamed it to `_no_delim` instead of `_flat`

src/lib.rs
qwandor: It looks like it isn't used anymore anyway.

src/winrtble/utils.rs
qwandor: I've added them back.

src/api/bdaddr.rs
qwandor: I think that's better. Done this myself.

PR #88, opened 2020-10-09
Use thiserror instead of failure

kjellkongsvik:


qdot:
Why was this closed? I hadn't had a chance to look at it yet.

PR #90, opened 2020-10-28
new example to use btleplug

sum12:
This example shows how to use the subscription mechanism.
In the example
- connect to the desired device (eg: mi band 5) using its bluetooth
address
- subscribe to the auth characteristic using a callback, and initiates
authentication process
- the callback pushes the received value (ValueNotification) in a channel
- this channel is consumed in a separate thread.
- this background threads handles the rest of communication

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,3 +56,4 @@ cocoa = "0.24.0"
 
 [dev-dependencies]
 rand = "0.7.3"
+openssl = "0.10.30"
--- a/examples/bandbridge.rs
+++ b/examples/bandbridge.rs
@@ -1,0 +1,151 @@
+use openssl::symm::{Cipher, Crypter, Mode};
+use std::str::FromStr;
+use std::sync::mpsc;
+use std::sync::mpsc::RecvTimeoutError;
+use std::thread;
+use std::time::Duration;
+
+use btleplug::api::{Central, Peripheral, ValueNotification, UUID};
+#[cfg(target_os = "linux")]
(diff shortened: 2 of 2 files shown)


sum12:
Is there anything I can do to move this ahead. This could be a pretty good example of the subscribe functionality.

qdot:
Ok, finally looking at this.

This seems... really device specific, without any real explanation in the example itself of what is happening? Also, bringing in openssl means compiling on windows gets really difficult due to having to vendor the native library in. I definitely like what it's trying to do, but I'm not really sure this is going to fit as a general example.

sum12:
Thanks for looking and agreed to those points. The openssl thing did not struck me neither that there is no explanation. I could rewrite it in a more explanatory way. However for communicating with miband5, openssl is kind of mandatory. 
Do you have better example to demonstrate the functionality? I will be happy to give it a try.

blandger:
Example seem quite interesting. Is there a way to handle disconnects by reconnecting BLE device ?

qwandor:
Is it possible to use rustls rather than openssl? That might be easier to get working on Windows.

sum12:
I could try to make it work with https://github.com/RustCrypto/block-ciphers and add documentation around the example) but not sure this can be solved 
> I'm not really sure this is going to fit as a general example.

qdot:
0.8.0 is releasing this week, which will invalidate the API for this, and we still haven't really hashed out whether examples for specific devices that may not be universally acceptable are really worth it. Closing.

PR #92, opened 2020-11-11
CoreBluetooth async Disconnect, Pending Connect, Discover Characteristics

trueb2:
In CoreBluetooth, disconnecting from a peripheral must not remove the Peripheral in order to set a pending connection. Since we don't have the whole of CoreBluetooth's async callbacks, we must separate and loosen some of the synchronous APIs that are exposed by the current Rust Peripheral abstraction. I am not saying that there shouldn't be nice synchronous abstractions, just that the current level prevents some use cases that I would consider to be standard.

I have the following use case that is supported on macOS by these changes.
* Begin scanning indefinitely with async event receiver
* Filter discovered peripherals by name
* Connect to a peripheral by bd_addr
* Set callbacks for the discovery of specific characteristics that use synchronous `api::Peripheral` methods `read` and `write`
* Initiate disconnection from the peripheral after a period of time
* Initiate a pending connection to the peripheral after another period of time
* Re-connect and repeat starting at setting callbacks

I suspect that the windows and Linux variants of this interface are invalidated by this change but have not verified.
- [ ] Don't break windows
- [ ] Don't break linux
- [ ] Gracefully handle remote peripheral disconnection
- [ ] Document api change

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,7 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.148.1/containers/rust/.devcontainer/base.Dockerfile
+
+FROM mcr.microsoft.com/vscode/devcontainers/rust:0-1
+
+# [Optional] Uncomment this section to install additional packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,34 @@
+{
+	"name": "Rust",
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+	"runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
+
+	// Set *default* container specific settings.json values on container create.
+	"settings": { 
+		"terminal.integrated.shell.linux": "/bin/bash",
+		"lldb.executable": "/usr/bin/lldb",
+		// VS Code don't watch files under ./target
+		"files.watcherExclude": {
+			"**/target/**": true
+		}
+	},
+
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"rust-lang.rust",
+		"bungcip.better-toml",
+		"vadimcn.vscode-lldb",
+		"mutantdino.resourcemonitor"
+	],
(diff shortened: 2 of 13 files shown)


trueb2:
I added the `.devcontainer` configs to build for Linux from mac in VS code + docker. After marking on_discovery `unimplemented!()`, the Linux variant builds but has slightly different call patterns for the event discovery example. I will update the windows on_discovery as well, but I have no way to build windows.

qdot:
So, hello from 3 months later. >.>

Just curious, was this ready to land? Looks like you at least got it compiling across all of our supported platforms.

trueb2:
Hi, it has been a bit since I looked at this as well. I had it passing linux, windows, and macos. 

I would like to revisit this when I get a chance. Honestly, I think it would be good to reduce the scope of the change. For example, the .devcontainer stuff should be separate but was used to verify builds on linux. Addditionally, I haven't looked at how the API compares with linux and windows.

It would be great to mirror the macOS CoreBluetooth, DispatchQueue, ReSwift, NIO models using the robust, readable, and performant Rust async patterns. So even though this works, is it in the direction that you want breaking changes to go?

There are conflicts now anyways preventing merge.

qdot:
> It would be great to mirror the macOS CoreBluetooth, DispatchQueue, ReSwift, NIO models using the robust, readable, and performant Rust async patterns. So even though this works, is it in the direction that you want breaking changes to go?

So yeah, I'm talking about that with @qwandor @NZSmartie now. Since Bluez is moved over to dbus and therefore a little easier to work with now, the next goal is to move this to async Rust. Right now the goal is to cut all of the cruft out of the library (there's so much either non-or-half implemented things), then see what's left, what needs to be replicated, etc. If you're interested in discussing this, I've got a discord channel for this project at https://discord.gg/QGhMFzR, though we'll also be discussing some of these ideas in issues here. 

I'd love to have more input on the CoreBluetooth side, as I basically cargo-culted (heh) what's currently available for the WebBluetooth implementation in Servo. Ideally I'd like to have this library able to work on both macOS and iOS too, and I have no idea what, if anything, needs to happen for that.

qwandor:
@trueb2 I've uploaded #114 with a first cut at what an async API might look like. I'm only really familiar with BlueZ on Linux though, so I'd be interested to hear how well you think it fits the CoreBluetooth model and what you have in mind.

qdot:
This is bitrotted at this point, since 0.8.0 will be out this week, so I'm closing. That said, we should at least revisit this and maybe break out things that are still missing in the 0.8.0 branch though.

PR #94, opened 2020-11-13
Propagate watcher errors on Windows

### LINKED ISSUE #91: Unwrapped result does not propagate the error
**Describe the bug**
The library calls unwrap on an error result.
```rust
if adapter.start_scan().is_err() {  // <-- HERE on start_scan()
    println!("Please enable bluetooth");
    exit(-1);
}
```
Bescause of this in [watcher.rs](https://github.com/deviceplug/btleplug/blob/master/src/winrtble/ble/watcher.rs) : 
```rust
pub fn start(&self, on_received: AdvertismentEventHandler) -> Result<()> {
        self.watcher
            .set_scanning_mode(BluetoothLEScanningMode::Active)
            .unwrap();
        let handler = TypedEventHandler::new(
            move |_sender, args: *mut BluetoothLEAdvertisementReceivedEventArgs| {
                let args = unsafe { &*args };
                on_received(args);
                Ok(())
            },
        );
        self.watcher.add_received(&handler).unwrap();
        self.watcher.start().unwrap();  // <-- THERE
        Ok(())
}
```

**Expected behavior**
The error should be returned and not brutally unwrapped so I can handle it gracefully.

**Actual behavior**
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: HRESULT 0x800710DF', [~\.cargo\registry\src\github.com-1ecc6299db9ec823\btleplug-0.5.4\src\winrtble\ble\watcher.rs:48:3](https://github.com/deviceplug/btleplug/blob/master/src/winrtble/ble/watcher.rs#L48)

**Additional context**
I'm on Windows 10, and my bluetooth adapter is disabled in the command center.
image
Propagating this error allows me to handle this case gracefully.

aklajnert:
Fixes #91 - I also got bitten by this, so I've decided to try to tackle it.  
I'm not sure if I've picked the right error type, I can't see which one is raised by the Linux equivalent.

For me, if I have the Bluetooth adapter turned off it raises `Error::Other(HRESULT 0x800710DF)`. Definitely, that's better than panicking.

### DIFF
--- a/src/winrtble/ble/watcher.rs
+++ b/src/winrtble/ble/watcher.rs
@@ -11,7 +11,7 @@
 //
 // Copyright (c) 2014 The Rust Project Developers
 
-use crate::Result;
+use crate::{Error, Result};
 use winrt::{
     windows::devices::bluetooth::advertisement::*, windows::foundation::TypedEventHandler, ComPtr,
     RtDefaultConstructible,
@@ -25,6 +25,12 @@ pub struct BLEWatcher {
 
 unsafe impl Send for BLEWatcher {}
 unsafe impl Sync for BLEWatcher {}
+
+impl From<winrt::Error> for Error {
+    fn from(err: winrt::Error) -> Error {
+        Error::Other(format!("{:?}", err))
+    }
+}
 
 impl BLEWatcher {
     pub fn new() -> Self {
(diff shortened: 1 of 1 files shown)


qdot:
In on dev at 3772aef, will be part of the 0.6 release. Had to make a few changes to accommodate the winrt library update. Thanks for the patch!

PR #95, opened 2020-11-23
Make Peripheral a struct to make the API more ergonomic

philipstears:
Hi there! Firstly, let me just say thanks for putting in the effort to try to pull together all of the abandoned bluetooth projects for rust. I've tried a few times to work with bluetooth in rust, and there have always been issues that have caused me to just throw my hands up and walk away for a bit. This project looks like the first that will eventually meet my needs, which are mostly home and brew automation focussed. But I digress.

This PR is here to serve as an invitation for discussion. 

A pain point that I, and seemingly others have had, is working with the Peripheral trait. It requires a deeper understanding of rust than is perhaps required. To use Peripheral in my projects, I need to box it, and that can't be done directly because Peripheral requires Clone, which disallows direct boxing. Instead, Peripheral needs wrapping up in a secondary trait object which doesn't require Clone. That raises the barrier to entry quite a bit, in my opinion.

I think the library could be made much easier to use by just removing the use of the trait, and following the pattern for cross-platform support that seems to be common with the standard library itself - using a tuple struct which contains the platform-specific type. 

This PR is a proof-of-concept of that approach, for the bluez code. Note that the Windows and macOS implementations won't compile with this PR, and I've not updated any of the documentation. I'm really just interested in hearing others' opinions on this approach.

If it's something that would be acceptable to be merged, I'll happily get this running under Windows and macOS, and sort out the documentation too.

### DIFF
--- a/src/api/adapter_manager.rs
+++ b/src/api/adapter_manager.rs
@@ -17,11 +17,8 @@ use std::sync::mpsc::{channel, Receiver, Sender};
 use std::sync::{Arc, Mutex};
 
 #[derive(Clone, Debug)]
-pub struct AdapterManager<PeripheralType>
-where
-    PeripheralType: Peripheral,
-{
-    peripherals: Arc<DashMap<BDAddr, PeripheralType>>,
+pub struct AdapterManager {
+    peripherals: Arc<DashMap<BDAddr, Peripheral>>,
 
     // Sender is never handled mutably, but mpsc's Sender is Send only, not Sync,
     // so we can't just wrap it in an Arc to pass it around as part of the adapter
@@ -41,10 +38,7 @@ where
     event_receiver: Arc<Mutex<Option<Receiver<CentralEvent>>>>,
 }
 
-impl<PeripheralType> AdapterManager<PeripheralType>
-where
-    PeripheralType: Peripheral + 'static,
-{
+impl AdapterManager {
     pub fn new() -> Self {
         let peripherals = Arc::new(DashMap::new());
         let (event_sender, event_receiver) = channel();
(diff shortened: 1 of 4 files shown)


olefriis:
> Instead, Peripheral needs wrapping up in a secondary trait object which doesn't require Clone.

Out of curiosity, how would you do that? I'm having the exact same problem now and have no idea how to do it, meaning all code using `Peripheral`s will get pushed into big, ugly functions instead of being nicely factored.

Just thinking that documenting how to work with `Peripheral`s would be a much lower-hanging fruit than fixing the code base for all supported platforms at once. Though I would love to help out with the changes if somebody decides that this is the way we want to go.

qdot:
Note that the whole API is being overhauled in #114, so you might want to check there for how things will look in the future (and probably not worry too much about trying to fix 0.6.x)

philipstears:
That's very exciting indeed, I can't wait to see how it shapes up. 

@olefriis - here's what I've done to work-around the Peripheral trait in my codebase:

https://github.com/philipstears/brew-monitor/blob/recipe-bridge/lib/bm-grainfather/src/bluetooth/btleplug.rs#L28

I declare my own trait (`ClientImpl`), with the API I want, and a struct that implements that trait on-top of `Peripheral`, and then I can box that trait normally (https://github.com/philipstears/brew-monitor/blob/recipe-bridge/lib/bm-grainfather/src/bluetooth/btleplug.rs#L87)

olefriis:
@philipstears Thanks a bunch - I'm still trying to wrap my head around this, but I'll get there 😄 Doing Rust, I need to learn all the time, which for me is one of the main attractions of the language (only using it on spare-time projects...).

@qdot That's very nice to hear! I must admit that I didn't know what to expect after reading #100, but super happy to hear that stuff is happening!

NZSmartie:
With the recent merge of #103, the platform specific `Peripheral` struct is now public and exposed directly to the library consumer. Along the use of _static_assertions_ to ensure the struct conforms to the `btleplug::api::Peripheral` trait. 

With that said, please let us know if there are any other concerns that are still to be addressed

PR #96, opened 2020-11-26
Store services on PeripheralProperties

philipstears:
In my use case (brew / home automation) I would like to be able to be monitoring devices as they become available, and do different things based on their advertised data. One of the facets of that data that is useful to act on, is the service id exposed by the device.

I saw a previous PR (https://github.com/deviceplug/btleplug/pull/23) that attempted to make more information available, but that was limited to bluez only.

I *think* the way things are intended to work, is that the `DeviceDiscovered` event just gives the address of the discovered advice, and then if someone wants more details, they can get the device, and use its `PeripheralProperties` to access the advertising data, therefore, this PR just extends PeripheralProperties with a BTreeSet of services, and (currently only for bluez) populates it when advertising information is available.

I've also changed the discovery example to show retrieving the device based on the address.

If this is the right approach, I can look at adding similar functionality for Windows and macOS too.

### DIFF
--- a/examples/event_driven_discovery.rs
+++ b/examples/event_driven_discovery.rs
@@ -52,7 +52,12 @@ pub fn main() {
     while let Ok(event) = event_receiver.recv() {
         match event {
             CentralEvent::DeviceDiscovered(bd_addr) => {
-                println!("DeviceDiscovered: {:?}", bd_addr);
+                println!(
+                    "DeviceDiscovered: {:?}",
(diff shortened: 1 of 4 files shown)


NZSmartie:
Ah woops, this got implemented in #126, though as a `Vec<Uuid>` instead of a `BTreeSet<Uuid>` if that is still important, then we could have it changed

PR #98, opened 2020-12-05
feat: Support winrt-rs v0.7

kaz399:
I modified winrtble code to support winrt-rs v0.7.

Since winrt-rs v0.7 is implemented by Microsoft. With this, the API is no longer compatible with previous versions.
I have confirmed that the code in the example works, but I have not been able to verify all the features.

Any comments are welcome.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,7 @@ version = "^4"
 features = ["verbose-errors"]
 
 [dependencies.winrt]
-# This is stuck at 0.6 until we update the UWP core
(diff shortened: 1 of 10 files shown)


qdot:
This is now on the dev branch at fdd226961d5a87460768fa5dce1b32bec467ef46. I had to rebase it on top of some other patches that came in before it.

My initial smoke tests with my library seem positive. Everything still connects and works in the way it should, subscriptions for notifications are still fine, etc.

Thanks again for doing this work!

PR #101, opened 2020-12-12
Store service data on PeripheralProperties

icewind1991:
Spiritual successor (for my use case) of #23, inspired by #96.

Currently only implemented for bluez, but the approach should be easy to implement for the other backends.


NZSmartie:
Implemented in #126

PR #102, opened 2020-12-12
move Peripheral from general parameter to associated type for Central

icewind1991:
With both `Adapter` and `Peripheral` being platform specific, I can't see any situation where you would want to implement `Central` more than once for a single `Adapter` but with different `Peripheral`s.

By removing the generic parameter instead it makes the trait easier to work with in generics.


qwandor:
I think this is a good idea. Can you rebase it against the latest `dev` branch? Or I'll do so after the async change is in.

qwandor (approved):

qwandor:
Never mind, done.

PR #103, opened 2020-12-12
unify api usage between platforms a bit

icewind1991:
By re-exporting the platform specific `Manager` and `Adapter` saves the user of the api from doing so and should prevent most api users from "accidentally" making their code platform specific.

By type-aliasing `Adapter `as `ConnectedAdapter` for mac and windows, and adding a no-op `connect` the user doesn't have to care that only linux
has a difference between the two.

### DIFF
--- a/src/bluez/adapter/mod.rs
+++ b/src/bluez/adapter/mod.rs
@@ -201,7 +201,7 @@ impl Adapter {
         );
     }
 
-    pub fn proxy(&self) -> Proxy<&SyncConnection> {
+    pub(crate) fn proxy(&self) -> Proxy<&SyncConnection> {
         self.connection
             .with_proxy(BLUEZ_DEST, &self.path, DEFAULT_TIMEOUT)
     }
--- a/src/bluez/adapter/mod.rs
+++ b/src/bluez/adapter/mod.rs
@@ -13,7 +13,7 @@
 
 // mod acl_stream;
 // mod dbus;
-mod peripheral;
+pub mod peripheral;
 
 use super::{
     bluez_dbus::adapter::OrgBluezAdapter1, bluez_dbus::device::OrgBluezDevice1Properties,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,6 +93,16 @@ pub mod corebluetooth;
 #[cfg(target_os = "windows")]
 pub mod winrtble;
 
+#[allow(unused_imports)]
+#[cfg(target_os = "linux")]
+pub use crate::bluez::{adapter::Adapter, manager::Manager};
+#[allow(unused_imports)]
+#[cfg(target_os = "macos")]
+pub use crate::corebluetooth::{adapter::Adapter, manager::Manager};
+#[allow(unused_imports)]
+#[cfg(target_os = "windows")]
+pub use crate::winrtble::{adapter::Adapter, manager::Manager};
+
 #[derive(Debug, thiserror::Error, Clone)]
 pub enum Error {
     #[error("Permission denied")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,19 +23,11 @@
 //! use std::thread;
 //! use std::time::Duration;
 //! use rand::{Rng, thread_rng};
-//! #[cfg(target_os = "linux")]
-//! use btleplug::bluez::{adapter::Adapter, manager::Manager};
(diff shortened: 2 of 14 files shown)


icewind1991:
Note that in order to get a consistent `connect` api I had to change the type signature for the bluez version, making this a breaking change. (Although I don't think it's likely to break anything in the wild)

icewind1991:
I have no need nor intention to maintain a fork, #101 (or something that achieves the same result) is the only ones I actually "need", the others were changes made after trying to make my code platform agnostic and noticing the annoyances of doing so with the current api.

If you prefer not to have more "structural" changes like this or #102 submitted without discussion first then I understand that.

Either way, if you don't like the changes from this or #102 or want to discuss their approaches first then that's fine by me.

icewind1991:
Reading around some more I can see that you aren't looking for for contributions that aren't well vetted and properly supported on all platforms because of the extra burden it puts on you as a developer, which is totally understandable.

I'm completely fine with closing these last two PR's and just leaving them as suggestions for future api stuff, I can easily pull the relevant bits into my own crate.

As for #101, I hope that it being a graceful fallback (any platform where it doesn't get implemented will just not have the data set, but nothing should break at least) makes it easier to merge. I can look into the windows implementation for it to if needed but, as is probably acommon issue, I don't have mac hardware to work on a mac implementation.

Either way, I thank you for taking up the task of creating this bluetooth library and hope that at least some of my work is useful.

qdot:
So just a heads up, after some people engaging on the Bluez side (which was going to be the biggest hurdle for me bringing up the new API, since the bluez socket stuff was rather cumbersome), I'm going to release v0.6 here in the next few days, then we're actually getting to that mythical point of working on API updates. Discussion is mostly taking place in issues here and/or on my discord (https://discord.gg/ah9W8WE). I'm leaving your PRs here so we can take some of your ideas into account too.

src/lib.rs
```diff
@@ -141,6 +141,16 @@ pub mod corebluetooth;
...
+#[allow(unused_imports)]
+#[cfg(target_os = "linux")]
+pub use crate::bluez::{manager::Manager, adapter::{ConnectedAdapter, Adapter}};
+#[allow(unused_imports)]
+#[cfg(target_os = "macos")]
+pub use crate::corebluetooth::{manager::Manager, adapter::{Adapter, Adapter as ConnectedAdapter}};
+#[allow(unused_imports)]
+#[cfg(target_os = "windows")]
+pub use crate::winrtble::{manager::Manager, adapter::{Adapter, Adapter as ConnectedAdapter}};
+
```
qwandor, line 153: Rather than putting this at the top level, how about putting the re-exports under a new module, called 'platform' or something like that?

This module could also have some tests to ensure that the various implementations implement the interfaces we expect them to (both the interfaces under api, and things like `Debug` or `Clone` or whatever else we expect).

src/bluez/adapter/mod.rs
```diff
@@ -521,7 +521,7 @@ impl Adapter {
         self.states.contains(&AdapterState::Up)
     }
 
-    pub fn connect(&self) -> Result<ConnectedAdapter> {
-        ConnectedAdapter::new(self)
+    pub fn connect(self) -> Result<ConnectedAdapter> {
+        ConnectedAdapter::new(&self)
     }
```
qwandor, line 526: If you rebase this against the latest dev branch the connect method shouldn't be necessary anymore.

src/lib.rs
```diff
@@ -141,6 +141,16 @@ pub mod corebluetooth;
 #[cfg(target_os = "windows")]
 pub mod winrtble;
 
+#[allow(unused_imports)]
+#[cfg(target_os = "linux")]
+pub use crate::bluez::{manager::Manager, adapter::{ConnectedAdapter, Adapter}};
+#[allow(unused_imports)]
+#[cfg(target_os = "macos")]
```
qwandor, line 148: or "ios"

qwandor:
I ended up rebasing this and fixing my own comments, plus adding `Peripheral` and reducing the visibility of a bunch of internal methods. Tell me what you think!

icewind1991:
Looks good, thanks for taking care of my PRs this week!

NZSmartie (approved):
This looks really tidy. I had to rebase on top of latest dev branch. before merging. just to verify changes.

Issue #52, opened 2020-07-13
Unable to scan for BTLE devices with Intel 5.1 BT adapter


pretty-gay-tho:
**Describe the bug**
btleplug - via intiface-cli-rs - is unable to scan for devices using the internal, Intel BT 5.1 adapter on my laptop.

**Expected behavior**
Devices should have been discovered.

**Actual behavior**
Devices were not discovered.

**Additional context**

While trying to get intiface-desktop working with my laptop and devices, I noticed that intiface-desktop was unable to load a list of devices on my system using an internal BT adapter. After debugging in Discord, we determined that a spare BT USB dongle worked perfectly. However, I can manually scan for BTLE devices with the non-working adapter via `bluetoothctl`.

Because we were ultimately able to reproduce this with intiface-cli-rs (which uses btleplug), it was decided that the best place for this report would be here.

### System info

OS: Arch Linux
`uname -a`: `Linux janeway 5.7.7-arch1-1 #1 SMP PREEMPT Wed, 01 Jul 2020 14:53:16 +0000 x86_64 GNU/Linux`
bluez version: 5.54 (actually built from [git HEAD](https://git.kernel.org/pub/scm/bluetooth/bluez.git/commit/?id=3ec6f74dc95cef929081be90855a104e11130150))
Non-working BT adapter info (internal BT adapter):
```
hci0:   Type: Primary  Bus: USB
        BD Address: 04:ED:33:10:1E:85  ACL MTU: 1021:4  SCO MTU: 96:6
        UP RUNNING PSCAN ISCAN
        RX bytes:24176 acl:0 sco:0 events:907 errors:0
        TX bytes:7014 acl:0 sco:0 commands:509 errors:0
        Features: 0xbf 0xfe 0x0f 0xfe 0xdb 0xff 0x7b 0x87
        Packet type: DM1 DM3 DM5 DH1 DH3 DH5 HV1 HV2 HV3
        Link policy: RSWITCH SNIFF
        Link mode: SLAVE ACCEPT
        Name: 'janeway'
        Class: 0x10010c
        Service Classes: Object Transfer
        Device Class: Computer, Laptop
        HCI Version: 5.1 (0xa)  Revision: 0x100
        LMP Version: 5.1 (0xa)  Subversion: 0x100
        Manufacturer: Intel Corp. (2)
```

Working BT adapter info (USB dongle):
```
hci1:   Type: Primary  Bus: USB
[...]

Issue #69, opened 2020-08-01
Consider supporting blurz or other dbus interface to bluetoothd


rvolgers:
btleblug currently takes the approach of directly using a hci socket to communicate with the kernel's bluetooth stack on Linux. An alternative approach is to use dbus to communicate with bluetoothd. It's probably worth at least discussing the trade-offs, and potentially adding support for the alternative approach.

Advantages of dbus:
* Much, much, *much* simpler, pretty much all logic is implemented in existing libraries instead of getting deep into HCI packet parsing in btleplug itself.
* (TODO: verify this) it seems like the dbus API might be more reliable, possibly because it caches information

Disadvantages:
* Dependency on separate daemon and dbus. I'm not sure where it is installed or not. It was present on my raspberry pi.
* If you use libdbus, cross compiling becomes harder (you need a custom docker container for cross). You could of course use a pure Rust dbus library (zbus), but that may cause longer compile times.

rvolgers:
Um, I don't remember *assigning* this to anyone, but I assume it's just how new issues work...

qdot:
First off, yup, everything here auto-assigns to me, just means I have less of a chance of losing track of it.

As for DBus, heh, I just closed #9 because I figured async'ing the sockets would just be a matter of using a generic async fd type. That said, since the system is extensible, it'd definitely be worth exploring this if someone feels up to it. I am admittedly weakest on bluez (both socket and dbus), and my very vague recollection was that the dbus API was missing... something? But this is from years ago, so if someone that's up on the current state knows more, would love to have more info.

oxc:
The DBus API can not, for example, explicitly initiate a passive scan. You can use the mgmt-api to add devices to the drivers whitelist, for which the kernel will scan passively. In both APIs, there is no such thing as "starting a passive scan".

The mgmt-api is not available via dbus, but requires opening a HCI (control) socket.

It might still be worth combining those two APIs, since the HCI control socket should be much easier to handle than the raw sockets.

NZSmartie:
I'm currently making progress on this, using `dbus-rs`. I've got adapter and peripheral discovery working 
https://github.com/NZSmartie/btleplug/tree/feature/bluez-dbus

NZSmartie:
as for the management socket, that still requires a setcap admin capability. Looking at the bluez source code for 5.54, passive scanning does seem to happen, but it is not explicit. I'll poke around with it a bit more once i've got my fork of the bluez APIs sorted.

[...]

PR #105, opened 2021-01-10
BlueZ: Replace raw HCI sockets with BlueZ over DBus

### LINKED ISSUE #60: DeviceLost event is never fired.
**Describe the bug**
I did small modification by adding  device lost event processing
            CentralEvent::DeviceLost(bd_addr) => {
                println!("DeviceLost: {:?}", bd_addr);

and that looks like program prints 'Discovered event' only after turning on BLE device.
Is 'DeviceLost' event not fired? I see description in comments:
> The first time the getter/ is called, it will return Some(Receiver<CentralEvent>). After that, it  will only return None

If that example does not works properly so probably it's worth deleting?

**Expected behavior**
I expected to see correct Discovered/Lost events processed for turned on/off device

**Actual behavior**
Program prints 'Discovered event' only after turning on BLE device.
When I try turn on/off device several times on Linux, it doesn't work, only one event if fired.

**Additional context**
Linux Mint 20, x64

NZSmartie:
This is a complete rewrite of the "bluez" platform specific code.

A lot has happened and I did my best to commend code to explain how/why a well as keep commits subjective so that changes can be understood in context.

Things I want to do before merging:
- [ ] ~~Verify I didn't break the existing API~~ (Maintaining a backwards compatible API is a lot harder than I thought and ends up with a lot of messy dead code)
- [ ] Ensure both passive and active discovery work as intended
- [x] Ensure the code isn't too terrible looking
- [ ] Test this library against multiple versions of BlueZ to see what fails where

Closes #60, #69


NZSmartie:
I had a look through other open issues and found two related issues that will be resolved with this PR

NZSmartie:
@qdot, can you add `libdbus-1-dev` to Linux's installed packages on the CI pipeline?

NZSmartie:
I've done what I can to test this PR, and I'm now relying on others to help test it. Since it works for me™, I'm happy to submit future PRs to fix bugs or other issues that I've yet to face

NZSmartie:
just rebased on to master

lnicola:
Also closes #52.

qdot:
Awesome, thanks for the updates. I am definitely trying to get to reviewing this, hopefully in the next day or two.

qdot:
(Note: This is permission to yell at me in these here comments if I don't get to it soon :3 )

src/bluez/adapter/mod.rs
```diff
@@ -11,517 +11,478 @@
...
 
-    pub fn is_up(&self) -> bool {
-        self.states.contains(&AdapterState::Up)
+    fn peripheral(&self, address: BDAddr) -> Option<Peripheral> {
+        self.manager.peripheral(address)
     }
 
-    pub fn connect(&self) -> Result<ConnectedAdapter> {
-        ConnectedAdapter::new(self)
+    fn active(&self, _enabled: bool) {
```
qdot, line 485: Here and elsewhere, can we get some comments on the methods that need to be filled in? Makes it easier to file help wanted bugs or at least say issues are claimed later, if we know what needs to be filled in.

src/bluez/adapter/peripheral.rs
```diff
@@ -11,410 +11,404 @@
...
+    connection: Arc<SyncConnection>,
+    path: String,
     address: BDAddr,
     properties: Arc<Mutex<PeripheralProperties>>,
     characteristics: Arc<Mutex<BTreeSet<Characteristic>>>,
-    stream: Arc<RwLock<Option<ACLStream>>>,
-    connection_tx: Arc<Mutex<Sender<u16>>>,
-    connection_rx: Arc<Mutex<Receiver<u16>>>,
-    message_queue: Arc<Mutex<VecDeque<ACLData>>>,
+    attributes_map: Arc<Mutex<HashMap<u16, (String, Handle, Characteristic)>>>,
```
qdot, line 58: If you wanted to lose the Mutex here, you could switch to DashMap. I use it heavily up in Buttplug, works great. That said, with everything else so lock heavy here, that's really just the lightest of nits and could be a followup.

src/bluez/adapter/peripheral.rs
```diff
@@ -11,410 +11,404 @@
...
+                self.address, manufacturer_data
+            );
+            let mut result = Vec::<u8>::new();
+            // dbus-rs doesn't really have a dictionary API... so need to iterate two at a time and make a key-value pair.
+            if let Some(mut iter) = manufacturer_data.0.as_iter() {
+                loop {
+                    if let (Some(id), Some(data)) = (iter.next(), iter.next()) {
+                        // This API is terrible.. why can't I just get an array out, why is it all wrapped in a Variant?
+                        let data: Vec<u8> = data
+                            .as_iter() // 🎶 The Variant is connected to the
```
qdot, line 318: I like this, please go through the rest of your patch and make your comments rhyme.

src/bluez/adapter/peripheral.rs
```diff
@@ -512,154 +472,126 @@ impl ApiPeripheral for Peripheral {
...
-            lock.insert(c.clone());
-        });
+        debug!("All services are now resolved!");
 
-        Ok(results)
+        Ok(self
+            .characteristics
+            .lock()
+            .unwrap()
+            .clone()
```
qdot, line 494: Very nit: You *might* be able to use cloned() after your map (and use refs up to that point), save the copy then discard? But it honestly doesn't matter much in this context, and I can't remember off the top of my head where cloned() comes in.

qdot (changes requested):
This is all looking really good! I've just got a few comments I'd like addressed, then it can land.

src/bluez/adapter/peripheral.rs
```diff
@@ -11,410 +11,404 @@
...
+                        // This API is terrible.. why can't I just get an array out, why is it all wrapped in a Variant?
+                        let data: Vec<u8> = data
+                            .as_iter() // 🎶 The Variant is connected to the
+                            .unwrap() // Array type!
+                            .next() // The Array type is connected to the
+                            .unwrap() // Array of integers!
+                            .as_iter() // Lets convert the
+                            .unwrap() // integers to a
+                            .map(|b| b.as_u64().unwrap() as u8) // array of bytes...
+                            .collect(); // I got too lazy to make it rhyme... 🎶
```
qwandor, line 325: You can do this more nicely with `dbus::arg::prop_cast`, or if you pass `--prop-newtype` to `dbus-codegen-rust` it will give you a type-safe wrapper for the `HashMap`, and you can do `OrgBluezDevice1Properties::manufacturer_data` to get a `HashMap<u16, Variant<Box<dyn RefArg>>>`, and then you can just map the values with `dbus::arg::cast::<Vec<u8>>(&v.0)`.

qdot:
@qwandor Thanks for the comment! 

If you or anyone else has additional guidance on the Bluez side, please feel free to add it here. I was mostly reviewing for structure, I'm far more familiar with the CoreBluetooth and UWP sides of the library, so I can't really comment much on the Bluez.

src/bluez/adapter/mod.rs
```diff
@@ -11,517 +11,478 @@
...
-                set.insert(f.clone());
+        // Spawn a new thread to process incoming DBus messages.
+        // This thread gets cleaned up in the 'impl Drop for Adapter'
+        *self.thread_handle.lock().unwrap() = Some(thread::spawn(move || {
+            while !should_stop.load(Ordering::Relaxed) {
+                // listener is protected by a mutex. as calling `process()` when a proxy is awaiting a reply will cause the reply to be dropped...
+                let lock = listener.lock();
+                while lock.process(Duration::from_secs(0)).unwrap() {}
+                drop(lock);
+                thread::sleep(Duration::from_millis(100));
```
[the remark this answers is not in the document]
NZSmartie: no source for this. I was going for a way to quickly process any messages there may be, and then sleeping on the thread for 100ms, so that other threads have a chance to use the listener to send/receive

src/bluez/adapter/peripheral.rs
NZSmartie: From memory, I gave this a try from the very start. but I think because i'm passing in a subset of the type that `dbus::arg::cast/prop_cast` are expecting, i get compile errors. I'll have another go and see if I've missed something.

qwandor: Ah, you may also need to make sure you have at least version 0.9.1 of the `dbus` crate, to include the fix from https://github.com/diwic/dbus-rs/pull/306.

src/bluez/adapter/mod.rs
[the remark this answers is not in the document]
NZSmartie: Looking back at my code and later learning about `parking_lot` I was dealing with mutex _fairness_, where, the `while` loop was constantly locking the mutex, not giving any other threads a chance to get a lock, and thus, anything that needed `listener` would block. 

I'll see if `parking_lot::ReentrantMutex` promotes fairness

src/bluez/adapter/peripheral.rs
qwandor: I've sent you https://github.com/NZSmartie/btleplug/pull/1 to fix this.

NZSmartie: Thanks @qwandor! Though i'm a bit sad to see my silly rhyming comment go away 😢

qdot: We need to keep some standards in this code, so I'm at least going to need some methods in iambic pentameter going forward.

qdot (approved):

qdot:
Ok, well, I'm cool with the changes and will land this, just gotta figure out why changing the base branch to dev borkt things because master and dev are on the same HEAD. I'd also like to keep my fast-forward only rule so I may rebase out the merge that comes in here. So, in short: Your code will be into the `dev` branch probably this evening, but may look slightly different than your commit lineup here, so you may need to reset if you had changes on your `dev`.

qdot:
Ok, I just brought this in with the merges for now, turns out the prehistory of rumble is just lousy with them anyways.

PR #108, opened 2021-01-23
Move from failure to thiserror

stygian-coffee:
This addresses issue #48. Failure is deprecated, and this makes errors implement `std::error::Error`, which is nice for using with things like `anyhow::Error`.


qdot:
Yaaaaaaaaaaay thanks! I love thiserror, I use it in a couple of other libraries. I need to bring in the bluez patch first (I don't *think* they'll overlap much I don't wanna screw up the other review I've got going), then I'll see about bringing this in.

qdot:
This is merged on dev as abe899f, I just had to make a conflict resolution 'cause the bluez stuff came in first.

qdot:
Thanks for the addition. :D

stygian-coffee:
Thanks for merging :)

PR #110, opened 2021-02-04
Remove unimplemented methods and fix return type of request

### LINKED ISSUE #104: What is Peripheral::request supposed to return?
GATT defines two ways of writing to a characteristic: write (ATT write request/response), and write without response (ATT write command). The 'response' in the in the write feature is just an acknowledgement that the write request was successful (or that it failed), and as far as I can see doesn't actually contain any data.

With that in mind, what is the `btleplug::api::Peripheral::request` supposed to return? Looking at the code, `request` isn't implemented at all for `winrtble`, and in the `corebluetooth` implementation just returns a copy of the value which was written. In the `bluez` implementation it seems to return something from BlueZ, but I can't figure out what.

It seems to me that this method should be changed to just return `Result<()>` like `Peripheral::command`. Or potentially, they could be combined into a single method, as the choice between using write or write without response is usually a matter of what the device supports rather than a conscious choice made by the application.

qwandor:
This removes `command_async`, `request_async`, `read_async` and `read_by_type_async`, which were not implemented on any platform. This is theoretically a breaking change, but any user of them is already broken because none of them actually did what they claimed to.

I've also removed the return value of `request`, because it seems to have originated from a misunderstanding of the GATT protocol. The device doesn't send a value back, only a confirmation that the write succeeded. On platforms where this was implemented it was just returning the same value which was written, which is not very useful.

We could consider going a step further and merging `request` and `command` into a single `write` method, as it is unlikely to matter to a user of the library whether it is really a write or a write-without-response. Any given characteristic will usually only support one or the other. The only platform which allows a distinction to be made between the two is Mac OS.

Fixes #104.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -142,8 +142,6 @@ pub struct ValueNotification {
 }
 
 pub type Callback<T> = Box<dyn Fn(Result<T>) + Send>;
-pub type CommandCallback = Callback<()>;
-pub type RequestCallback = Callback<Vec<u8>>;
 
 pub type NotificationHandler = Box<dyn FnMut(ValueNotification) + Send>;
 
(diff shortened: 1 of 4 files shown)


qdot:
Patch looks great, followup on the `write` question at #111

PR #112, opened 2021-02-05
Remove discover_characteristics_in_range.

qwandor:
It was either unimplemented or did exactly the same as discover_characteristics.


qdot:
Looks like this broke on linux, where something was actually using the _start/_end parameters (despite them being prepended with _ 😒)

qwandor:
Whoops, I thought I tested that. Fixed now.

qdot:
'sall good, I use CI as my linux tester constantly. >.>

PR #113, opened 2021-02-05
Replace request and command with a single write method.

### LINKED ISSUE #111: What do we want to do with Write vs. WriteWithResponse?
API bikeshedding time! Breaking out the discussion from #110 over here.

btleplug inherited its method names from rumble, which made some odd choices, I'm guessing either driven by bluez or the bluetooth spec. Request is basically WriteWithoutResponse, while Command I think is WriteWithResponse?

Anyways, in the new API, I think it'd be nice to collapse these to just write, as @qwandor mentioned. I think we *do* need to support both w/ and w/o response, possibly via a boolean argument, as multiple platforms can handle this capability (here's my old C# UWP code for handling that: https://github.com/buttplugio/buttplug-csharp/blob/master/Buttplug.Server.Managers.UWPBluetoothManager/UWPBluetoothDeviceInterface.cs#L296), and there are devices that expose both characteristic properties.

In my libraries that live on top of this, I have a full Write options object that can be built, but that's because I have to support Bluetooth/Serial/USB/god knows what else. This signature could probably just be `write(data: &[u8], write_with_response: bool)`? Or we could just have write/write_with_response methods (this might actually be clearer). If someone requests a type a characteristic doesn't handle, we throw an error.

qwandor:
This is hopefully easier to understand, and more similar to other GATT APIs.

It appears that the write types were backwards on Mac OS before, so this also fixes that.

I don't have a Windows machine to build on, so I haven't been able to implement this on Windows.

Fixes #111.

### DIFF
--- a/src/bluez/adapter/peripheral.rs
+++ b/src/bluez/adapter/peripheral.rs
@@ -23,7 +23,7 @@ use crate::{
     api::{
         AdapterManager, AddressType, BDAddr, CentralEvent, CharPropFlags, Characteristic,
         NotificationHandler, Peripheral as ApiPeripheral, PeripheralProperties, ValueNotification,
-        WriteKind, UUID,
+        WriteType, UUID,
     },
     bluez::{
         bluez_dbus::device::OrgBluezDevice1, bluez_dbus::device::OrgBluezDevice1Properties,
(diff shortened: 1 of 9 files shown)


qwandor:
I called the parameter `kind` and the enum `WriteKind` because `type` is a Rust keyword. Happy to rename to `write_type` / `WriteType` if you prefer, or if you have a better suggestion.

qwandor:
On further thought, I decided to rename them.

src/bluez/adapter/peripheral.rs
```diff
@@ -459,20 +460,30 @@ impl ApiPeripheral for Peripheral {
...
+                    WriteType::WithoutResponse => "command",
+                }
+                .to_string(),
+            )),
+        );
         Ok(self
             .proxy_for(&characteristic)
-            .map(|p| p.write_value(Vec::from(data), HashMap::new()))
+            .map(|p| p.write_value(Vec::from(data), options))
             .ok_or(Error::NotSupported("write_without_response".to_string()))??)
```
NZSmartie, line 483: Minor request
```patch
-.ok_or(Error::NotSupported("write_without_response".to_string()))??)
+.ok_or(Error::NotSupported("write".to_string()))??)
```

qwandor: Done.

PR #114, opened 2021-02-05
Make API async

qwandor:
This PR moves the API to an async model.

The Linux implementation is a fairly thin wrapper around my [bluez-async](https://crates.io/crates/bluez-async) crate, while the Mac OS implementation is a minimal modification of the existing Mac OS implementation, just turning the existing async blocks into functions.

The Linux implementation depends on Tokio because it ends up calling into `dbus-tokio`. There's no easy way around the `dbus-tokio` dependency, so I've also changed the Mac OS implementation to use Tokio rather than `async-std` for consistency.

The Windows implementation builds but is not actually async yet, pending https://github.com/microsoft/windows-rs/issues/375.

Feedback on the API design, implementations and how to test that it actually works properly would be very welcome.

TODO:
- [x] Implement on Mac OS.
- [x] Implement on Windows.
- [x] Add examples.


qwandor:
The diff will be easier to read once #115 is in, as this builds on top of it.

src/bluez/peripheral.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+}
+
+impl From<&CharacteristicInfo> for Characteristic {
+    fn from(characteristic: &CharacteristicInfo) -> Self {
+        Characteristic {
+            start_handle: 0,
+            end_handle: 0,
+            value_handle: 0,
```
NZSmartie, line 209: If we're not using these properties anywhere and the API is getting re-written. I think it's safe to drop these from `PeripheralProperties`

src/api/mod.rs
```diff
@@ -253,66 +252,59 @@ pub enum WriteType {
     WithoutResponse,
 }
 
+#[async_trait]
```
NZSmartie, line 255: nit: The attribute would it just above the pub trait declaration, instead of above the doc comments. to make it easier to identify in the event the documentation grows to be quite large

NZSmartie (approved):

src/api/mod.rs
qwandor: Done.

src/bluez/peripheral.rs
qwandor: Good point, done.

src/bluez/adapter.rs
```diff
@@ -0,0 +1,117 @@
...
+                ))
+            }
+        }
+        BluetoothEvent::Device {
+            id,
+            event: DeviceEvent::RSSI { rssi: _ },
+        } => {
+            let device = session.get_device_info(&id).await.ok()?;
+            Some(CentralEvent::DeviceUpdated((&device.mac_address).into()))
+        }
```
NZSmartie, line 99: The old behaviour of the BlueZ implementation would always emit a `CentralEvent::DeviceDiscovred` first before a `CentralEvent::DeviceUpdated` if a device was unknown to _btleplug_ at the time. This was done with a `AdapterManager<Peripheral>`. 
This implementation does not keep track of the state of known devices, thus emits `CentralEvent::DeviceUpdated` with no corresponding `CentralEvent::DeviceDiscovered`. This makes sense, since the device would already be known to BlueZ at the time, but not to _btleplug_. 

Since this is unspecified behaviour for _btleplug_, I'm not sure if it's required to address before merging, or if we should start writing up a spec and then have the platform implementations follow

qwandor: I've added some code to make the event stream always start with `DeviceDiscovered` events for every device known at that point. What do you think?

src/bluez/adapter.rs
```diff
@@ -0,0 +1,129 @@
...
+        BluetoothEvent::Device {
+            id,
+            event: DeviceEvent::ManufacturerData { manufacturer_data },
+        } => {
+            let device = session.get_device_info(&id).await.ok()?;
+            if manufacturer_data.len() > 1 {
+                warn!("Got more than one manufacturer data entry.")
+            }
```
NZSmartie, line 119: Would it be a better idea to emit an event for each manufacturer data that is received, instead of dropping them?

qwandor: Rebased on your PR, which solves this problem in a better way.

src/api/adapter_manager.rs
```diff
@@ -22,33 +27,15 @@ where
...
-
-    // Normally we'd just return the event receiver when an adapter is created.
-    // However, since adapters are cloned and retrieved via lists, this is really
-    // hard without changing the fundamentals of the API (which I want to do at
-    // some point, but not now). Storing an option here means that we'll only ever
-    // have one event receiver (as mpsc isn't clonable, which is what we want on
-    // the receiver side anyways), but means we also don't have to deal with the
-    // adapter API yet.
-    event_receiver: Arc<Mutex<Option<Receiver<CentralEvent>>>>,
+    async_senders: Arc<Mutex<Vec<UnboundedSender<CentralEvent>>>>,
```
qdot, line 30: Can we just make this a tokio Broadcast channel?

src/bluez/adapter.rs
```diff
@@ -0,0 +1,125 @@
...
+        // event than to miss one. It's unlikely to happen in any case.
+        let events = self.session.event_stream().await?;
+
+        // Synthesise `DeviceDiscovered' events for existing peripherals.
+        let devices = self.session.get_devices().await?;
+        let initial_events = stream::iter(
+            devices
+                .into_iter()
+                .map(|device| CentralEvent::DeviceDiscovered(BDAddr::from(&device.mac_address))),
+        );
```
qdot, line 37: I'm a little confused by this. We emit device discovered *every* time we get an event stream from the adapter?

src/bluez/adapter.rs
qdot: > if we should start writing up a spec and then have the platform implementations follow

You mean the step that this whole project was waiting on in the first place? :3

qdot (changes requested):
Ok, so, my major issue here is that *everything* is async. For connect/disconnect/characteristics/etc, that's totally fine, those are async operations across the board. But it seems really weird that event getters/setters (or just fetching event streams?!) are async. This is going to make the API really inflexible in terms of what can be called where in dependent libraries, and I'm worried it's all going to be in service of one platform, linux on bluez. All of the bluez calls go *all* the way back to bluez all the time, meaning we're always waiting on the system. Is there no way to cache things forward into the library?

Both CoreBluetooth and WinRT/UWP give us events we can listen for and update accordingly, and I'm just not familiar with what bluez gives us, but seeing there's an event stream, it seems like we should be able to do that there too. If that's the case, it's far more ergonomic to trust that event stream to be infallible and update our own properties that way, instead of forcing a system fetch on one platform and then making everyone else async even though they don't need it.

Filing this as request changes, but that's just because I wanna talk through this.

src/bluez/adapter.rs
NZSmartie: Would using an `btleplug::api::AdapterManager`  to track which devices are new, and which need to be updated instead? Much like how the [`winrt`](https://github.com/qwandor/btleplug/blob/688b9f2e661d6f72182ca04e0609e029e0f59952/src/winrtble/adapter.rs#L49-L60) and [`corebluetooth`](https://github.com/qwandor/btleplug/blob/688b9f2e661d6f72182ca04e0609e029e0f59952/src/corebluetooth/adapter.rs#L47-L69) (Although, admittidly, corebluetooth doesn't appear to emit `DeviceUpdated` events).

src/api/adapter_manager.rs
qwandor: Unfortunately `tokio_stream::wrappers::BroadcastStream<T>` implements `Stream<Item=Result<T, BroadcastStreamRecvError>>` rather than `Stream<Item=T>`, which would pollute our API and mean clients have to handle `BroadcastStreamRecvError`. It would also mean adding a dependency on `tokio_stream`, though that's less of an issue.

src/bluez/adapter.rs
qwandor: Currently yes. That's why I was asking on Discord exactly what behaviour people expect for devices which the system already knows about before `events()` is called. I still prefer leaving it up to the user to call `peripherals()` if they want to get existing devices, but this approach is more consistent with what the other platforms do. I assume calling `events()` multiple times is not how we expect people to use the API in most cases anyway.

qwandor:
> Ok, so, my major issue here is that _everything_ is async. For connect/disconnect/characteristics/etc, that's totally fine, those are async operations across the board. But it seems really weird that event getters/setters (or just fetching event streams?!) are async. This is going to make the API really inflexible in terms of what can be called where in dependent libraries, and I'm worried it's all going to be in service of one platform, linux on bluez. All of the bluez calls go _all_ the way back to bluez all the time, meaning we're always waiting on the system. Is there no way to cache things forward into the library?
> 
> Both CoreBluetooth and WinRT/UWP give us events we can listen for and update accordingly, and I'm just not familiar with what bluez gives us, but seeing there's an event stream, it seems like we should be able to do that there too. If that's the case, it's far more ergonomic to trust that event stream to be infallible and update our own properties that way, instead of forcing a system fetch on one platform and then making everyone else async even though they don't need it.

I haven't looked much at the Windows APIs yet, but this is definitely a difference between the CoreBluetooth and BlueZ models. The BlueZ model is that there is a persistent system-wide tree of objects (adapters, peripherals, services, characteristics, descriptors) which we can query and operate on. We can get events for changes to that tree (e.g. a new peripheral was added, some specific property of a peripheral changed), but they are at a higher level of abstraction than 'an advertisement was received'. And remember that the state is system-wide, so for example even if we haven't requested discovery, another application may have (or indeed might be connecting to devices and discovering services), and so things might be changing as a result of that.

We could maintain our own copy of all this within btleplug. But it would add a bunch of extra complexity, and unless we are very careful could easily get out of sync with definitive state that BlueZ maintains. It also seems wasteful, as we'd presumably need to have a background task maintaining our copy of the state at all times, even if the application using btleplug is doing nothing and doesn't care about what's going on, or just connects to one peripheral and doesn't care about all the others once it's found it. Fetching the peripherals directly from BlueZ when we actually need them seems like a better approach to me.

qdot:
> We could maintain our own copy of all this within btleplug. But it would add a bunch of extra complexity, and unless we are very careful could easily get out of sync with definitive state that BlueZ maintains. It also seems wasteful, as we'd presumably need to have a background task maintaining our copy of the state at all times, even if the application using btleplug is doing nothing and doesn't care about what's going on, or just connects to one peripheral and doesn't care about all the others once it's found it. Fetching the peripherals directly from BlueZ when we actually need them seems like a better approach to me.

WebBluetooth also doesn't do this, nor does Android. So we'd be hamstringing the ergonomics of rest of the platforms on Bluez and Bluez alone, which is exactly what I was going to try to get away from with the next version of the library, because this is exactly what Rumble did and it made things extremely awkward. This API will either force users of this library to make simple status fetches async all the way up through their system, or else force them to do the caching themselves to get around that if they're not using Bluez.

For me personally, I have thousands of users on windows/WASM/etc to worry about here, versus 10s on linux/Bluez, so that's going to reflect in my opinion here.

I'm also not sure how you see a background task as wasteful. If it's sitting there waiting for something that's already async, it's not really costing much. I live on event loops like that in a ton of my libraries. Rust async is built to handle 10ks of tasks, 1 waiting on some infrequent device updates doesn't seem like it should hurt, especially since there's already async bindings to dbus?

src/api/adapter_manager.rs
qdot: No, it won't. You use tokio_stream to rewrap the Stream, I do this in Buttplug already with no API pollution.

https://github.com/buttplugio/buttplug-rs/blob/master/buttplug/src/util/stream.rs#L5

qwandor:
> > We could maintain our own copy of all this within btleplug. But it would add a bunch of extra complexity, and unless we are very careful could easily get out of sync with definitive state that BlueZ maintains. It also seems wasteful, as we'd presumably need to have a background task maintaining our copy of the state at all times, even if the application using btleplug is doing nothing and doesn't care about what's going on, or just connects to one peripheral and doesn't care about all the others once it's found it. Fetching the peripherals directly from BlueZ when we actually need them seems like a better approach to me.
> 
> WebBluetooth also doesn't do this, nor does Android. So we'd be hamstringing the ergonomics of rest of the platforms on Bluez and Bluez alone, which is exactly what I was going to try to get away from with the next version of the library, because this is exactly what Rumble did and it made things extremely awkward. This API will either force users of this library to make simple status fetches async all the way up through their system, or else force them to do the caching themselves to get around that if they're not using Bluez.

Most of the other platform APIs don't have a way to get a list of peripherals (except perhaps certain subsets, such as paired peripherals), so having the `Central::peripherals()` method at all is arguably a choice to emulate something more like the BlueZ API. We could remove it and just have events, but that might make extra work for people migrating to the new version of btleplug.

If we are going to have a method like this to get a whole bunch of state from the system, I don't really see why it being async is a problem. If you're writing an app using an async model, then aren't most of your methods which do anything more than simple self-contained logic going to be async anyway? And if you're fetching the list of devices, then you've probably also enabled discovery (which is always going to be async, as it requires calling down to the system API) and are soon going to connect to a device, which also definitely needs to be async.

But perhaps I'm misunderstanding your concern. Which methods exactly are you concerned about? And where would you expect them to be used outside an async context?

> For me personally, I have thousands of users on windows/WASM/etc to worry about here, versus 10s on linux/Bluez, so that's going to reflect in my opinion here.
> 
> I'm also not sure how you see a background task as wasteful. If it's sitting there waiting for something that's already async, it's not really costing much. I live on event loops like that in a ton of my libraries. Rust async is built to handle 10ks of tasks, 1 waiting on some infrequent device updates doesn't seem like it should hurt, especially since there's already async bindings to dbus?

The problem is that if btleplug is to maintain it's only version of the peripheral list then it would need to be listening to all updates from all peripherals, even ones the application doesn't care about, but are getting updates because of some other application on the system requesting discovery, connecting and discovering services, potentially subscribing to characteristics, and so on.

qdot:
> Most of the other platform APIs don't have a way to get a list of peripherals (except perhaps certain subsets, such as paired peripherals), so having the `Central::peripherals()` method at all is arguably a choice to emulate something more like the BlueZ API. We could remove it and just have events, but that might make extra work for people migrating to the new version of btleplug.

I suppose this is a question of "how much state does this library maintain versus how much is the user required to maintain". Looking at it from that perspective, I'm not sure I have a good answer. I'm used to dealing with a very specific set of devices that I expect to be always connected (or else I don't really care about them if they aren't), and that's my single usage of the library. If we need to cover more usages than that, then it's probably out of the scope of what I'm thinking of here.
 
> If we are going to have a method like this to get a whole bunch of state from the system, I don't really see why it being async is a problem. If you're writing an app using an async model, then aren't most of your methods which do anything more than simple self-contained logic going to be async anyway? And if you're fetching the list of devices, then you've probably also enabled discovery (which is always going to be async, as it requires calling down to the system API) and are soon going to connect to a device, which also definitely needs to be async.

I suppose I expect state to be directly available, and actions to be async, but that's mostly because I'm thinking in event driven models at the moment. I see things like connection state as simple self-contained logic, but that may just be my usage. If the events ARE emitted, then I *can* deal with that above the library without having to make the async call specifically. Up until this rewrite everything had been a consensus of 1, so that may be reflected in how strong my opinions are here, but I may also just be gumming up the works and async getters for state might be fine.

It may just be best to make everything async to get this done, then dogfood that on beta a bit before releasing as v1?
 
> But perhaps I'm misunderstanding your concern. Which methods exactly are you concerned about? And where would you expect them to be used outside an async context?

In my libraries that depend on this, things like fetching connection status are sync, because I assume it's just state that should exist. But that once again comes with the assumption of persistent connection state (the "if it's not connected I don't care about it anymore" thing) which I think may differ from the full needs of a bluetooth library.
 
> The problem is that if btleplug is to maintain its own version of the peripheral list then it would need to be listening to all updates from all peripherals, even ones the application doesn't care about, but are getting updates because of some other application on the system requesting discovery, connecting and discovering services, potentially subscribing to characteristics, and so on.

That's a good point. I wasn't really thinking about how well this library would play with others.

I'll step back on my async split demands for now, and just play with this once it's working everywhere and dogfoodable to see if the thoughts I have on the ergonomics holds up, and we can go from there. 

So, that said, what's the review state of this at the moment seeing that the Windows repo has the async updates?

qwandor:
As far as I'm concerned it's ready to go. I've made the Windows implementation all async too, and it seemed to work as far as I tested it.

In terms of events, I think it would make sense to rethink how they work as well, and potentially have a way to get events from just a single device, or filter by a set of devices you are interested in, rather than always everything from all devices on the system. But that can be a separate PR.

src/api/adapter_manager.rs
qwandor: It looks like that stream will end if `Receiver::recv` returns `Err(RecvError::Lagged)`, which seems wrong.

qdot (approved):

qdot:
Ok, finally looked through everything again, r+'ing for now. Def interested in talking more about the event model, and all of the unwraps in windows have become a Problem lately (I'm seeing a lot more aborted operations after one of the latest windows patches, that end up panicking in an unwrap()), but it'll be good to at least start dogfooding this on dev to get an idea of how it's gonna feel.

qwandor:
Yeah, I'm definitely keen to improve the error handling, there are a bunch of unwraps all over the place that could return a `Result` instead.

PR #115, opened 2021-02-06
Clean up corebluetooth module

qwandor:
This cleans up several things on Mac OS:
* Adds a `Debug` implementation for many types.
* Removes some unused code and fixes some other warnings.
* Sets local_name to `None` rather than `Some("nil")` for devices without a name.
* Either logs or returns errors in some cases where they were being ignored (and causing compile warnings).

Let me know if you'd prefer it to be split into smaller PRs.


qdot:
We may not need async-std at all. That was a very nasty stop gap just to get things working. After I get a chance to review the bluez async work I may be able to clean that up.

PR #116, opened 2021-02-10
Embed Info.plist with NSBluetoothAlwaysUsageDescription

### LINKED ISSUE #106: prompt NSBluetoothAlwaysUsageDescription in macOS Big Sur
[access] This app has crashed because it attempted to access privacy-sensitive data without a usage description.  The app's Info.plist must contain an NSBluetoothAlwaysUsageDescription key with a string value explaining to the user how the app uses this data.

qwandor:
This is necessary to use Bluetooth on Mac OS 11.2.

Attempt to fix #106. Unfortunately this doesn't seem to work, any idea what I'm missing?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,9 @@ cocoa = "0.24.0"
 objc = "0.2.7"
 libc = "0.2.85"
 
+[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dev-dependencies]
+embed_plist = "1.2"
+
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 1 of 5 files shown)


qwandor:
Looks like the actual solution is to grant Bluetooth access to the terminal application.

qdot:
Setting terminal doesn't fix all cases. In my use case, I have another program that's running the executable outside of a shell context. For that, I've had to use the plist next to the executable, which works for me, while setting the terminal permission didn't do anything. So we probably need both listed in the README.

Looks like this is covered in the README PR, just want to leave a comment here saying *both* are valid options. :)

PR #121, opened 2021-02-14
Clean up dependencies

qwandor:
This removes some unnecessary dependencies, moves some to only the platforms that actually require them, and removes 2015 edition style `extern crate` declarations in favour of using macros individually.


NZSmartie (approved):
Looks Good To Me!

PR #123, opened 2021-02-19
Use uuid::Uuid in public API rather than defining our own UUID type.

qwandor:
A 16-bit UUID isn't a different kind of UUID, it's just a short encoding for a particular subset of UUIDs. This can be handled by an extension trait and a couple of helper functions.

These were copied from the `bluez_async` crate, but I wrote them and I'm happy for them to be under the btleplug license.

Does part of #68.


src/api/uuid.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    fn succinctly(&self) -> String {
+        if let Some(uuid16) = self.to_ble_u16() {
+            format!("{:#04x}", uuid16)
+        } else if let Some(uuid32) = self.to_ble_u32() {
+            format!("{:#06x}", uuid32)
+        } else {
+            self.to_string()
+        }
+    }
```
NZSmartie, line 59: I would like to see test cases for `succinctly`.
Also, I had to look up the dictionary meaning of _succinctly_. Since it's another flavour of `to_string()`, may I suggest renaming it to `to_string` and the `self.to_string()` be replaced with `ToString::to_string(self)` so that the appropriate fallback still gets called.

src/api/uuid.rs
```diff
@@ -0,0 +1,121 @@
...
+        if value & BLUETOOTH_BASE_MASK_16 == BLUETOOTH_BASE_UUID {
+            Some((value >> 96) as u16)
+        } else {
+            None
+        }
+    }
+
+    fn succinctly(&self) -> String {
+        if let Some(uuid16) = self.to_ble_u16() {
+            format!("{:#04x}", uuid16)
```
NZSmartie, line 53: Since the printed number may not be obviously hexadecimal, would prefixing them with `0x` be appropriate? A supporting argument is, in the official Bluetooth Assigned Numbers document, all the 16bit UUIDs are also prefixed with `0x`

src/api/uuid.rs
```diff
@@ -0,0 +1,121 @@
...
+/// UUID.
+pub const fn uuid_from_u32(short: u32) -> Uuid {
+    Uuid::from_u128(BLUETOOTH_BASE_UUID | ((short as u128) << 96))
+}
+
+/// Convert a 16-bit BLE short UUID to a full 128-bit UUID by filling in the standard Bluetooth Base
+/// UUID.
+pub const fn uuid_from_u16(short: u16) -> Uuid {
+    uuid_from_u32(short as u32)
+}
```
NZSmartie, line 17: Could these functions be part of the `BleUuid` trait? it feels more natural to have
```rust
use btleplug::api::uuid::BleUuid;
// ...
let my_32bit_uuid = Uuid::from_u32(...);
let my_16bit_uuid = Uuid::from_u16(...);
```
as opposed to
```rust
use btleplug::api::uuid::{uuid_from_u32, uuid_from_u16};
// ...
let my_32bit_uuid = uuid_from_u32(...);
let my_16bit_uuid = uuid_from_u16(...);
```

You would lose the `const fn`, but that is an unstable feature that may become available in the near future

src/api/uuid.rs
NZSmartie: Forgot to suggest adding this change for 32bit UUIDs as well

qwandor: This does prefix them with `0x`.

src/api/uuid.rs
qwandor: I agree that that would be nice, but I think that being a `const fn` is more important. In my experience so far I quite often want to have a bunch of constants up the top of a program like

```rust
const MY_CHARACTERISTIC: Uuid = uuid_from_u16(0x1234);
```

I'll add a TODO to make these part of the trait once that's allowed.

src/api/uuid.rs
qwandor: Added some tests.

I think having two functions on different traits implemented on the same type both called `to_string` is a recipe for confusion. I'm happy to call it something else though if you prefer. `to_short_string`? Any other suggestions?

NZSmartie: `to_short_string`, `to_compact_string`, `to_minimal_string`, uhh `to_🤏_string`

qwandor: Renamed.

NZSmartie (approved):

PR #125, opened 2021-02-21
Fix ordering of Characteristic display.

impala454:
Just a quick simple fix, I noticed the ordering was wrong in the Characteristic display function.


NZSmartie (approved):
Good catch, thank you for your contribution!

PR #126, opened 2021-02-22
[Feature] Advertisement support for Service Data, Manufacturer Data and Services

NZSmartie:
Bluetooth advertisement may also contain Service Data (that is, Service UUIDs followed by arbitrary bytes) and advertised services. 
This Merge Request adds 
1. `service_data` and `services` to `btleplug::api::PeripheralPropties` 
2. `ManufacturerDataAdvertisement`, `ServiceDataAdvertisement` and `ServiceSAdvertisement` to `btleplug::api::CentralEvent`
3. Emits appropriate events from BlueZ, WinRT and Core Bluetooth
   - WinRT requried some manual parsing as there isn't a direct API for Service Data
   - Core Bluetooth didn't implement manufacturer data, so that was also added

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -222,6 +222,11 @@ pub struct PeripheralProperties {
     /// Advertisement data specific to the device manufacturer. The keys of this map are
     /// 'manufacturer IDs', while the values are arbitrary data.
     pub manufacturer_data: HashMap<u16, Vec<u8>>,
+    /// Advertisement data specific to a service. The keys of this map are
+    /// 'Service UUIDs', while the values are arbitrary data.
+    pub service_data: HashMap<Uuid, Vec<u8>>,
+    /// Advertised services for this device
+    pub services: Vec<Uuid>,
     /// Number of times we've seen advertising reports for this device
     pub discovery_count: u32,
     /// True if we've discovered the device before
@@ -301,13 +306,30 @@ pub trait Peripheral: Send + Sync + Clone + Debug {
 }
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Clone)]
 pub enum CentralEvent {
     DeviceDiscovered(BDAddr),
     DeviceLost(BDAddr),
     DeviceUpdated(BDAddr),
     DeviceConnected(BDAddr),
     DeviceDisconnected(BDAddr),
+    /// Emitted when a Manufacturer Data advertisement has been received from a device
+    ManufacturerDataAdvertisement {
+        address: BDAddr,
+        manufacturer_id: u16,
+        data: Vec<u8>,
+    },
+    /// Emitted when a Service Data advertisement has been received from a device
+    ServiceDataAdvertisement {
+        address: BDAddr,
+        service: Uuid,
+        data: Vec<u8>,
+    },
+    /// Emitted when the advertised services for a device has been updated
+    ServicesAdvertisement {
+        address: BDAddr,
+        services: Vec<Uuid>,
+    },
 }
 
 /// Central is the "client" of BLE. It's able to scan for and establish connections to peripherals.
--- a/src/bluez/adapter/peripheral.rs
+++ b/src/bluez/adapter/peripheral.rs
@@ -237,6 +237,7 @@ impl Peripheral {
     pub fn update_properties(&self, args: OrgBluezDevice1Properties) {
         trace!("Updating peripheral properties");
         let mut properties = self.properties.lock().unwrap();
+        let mut emit_updated = false;
 
         properties.discovery_count += 1;
 
@@ -266,6 +267,7 @@ impl Peripheral {
         if let Some(name) = args.name() {
             debug!("Updating \"{}\" local name to \"{:?}\"", self.address, name);
             properties.local_name = Some(name.to_owned());
+            emit_updated = true;
         }
 
         if let Some(services_resolved) = args.services_resolved() {
@@ -281,18 +283,6 @@ impl Peripheral {
             cvar.notify_all();
         }
 
-        // if let Some(services) = args.get("ServiceData") {
-        //     debug!("Updating services to \"{:?}\"", services);
-
-        //     if let Some(mut iter) = services.0.as_iter() {
-        //         loop {
-        //             if let (Some(uuid), ())
-        //         }
-        //     }
-        // }
-
-        // As of writing this: ManufacturerData returns a 'Variant({<manufacturer_id>: Variant([<manufacturer_data>])})'.
-        // This Variant wrapped dictionary and array is difficult to navigate. So uh.. trust me, this works on my machine™.
         if let Some(manufacturer_data) = args.manufacturer_data() {
             debug!(
                 "Updating \"{}\" manufacturer data \"{:?}\"",
@@ -302,6 +292,12 @@ impl Peripheral {
                 .into_iter()
                 .filter_map(|(&k, v)| {
                     if let Some(v) = cast::<Vec<u8>>(&v.0) {
+                        self.adapter
+                            .emit(CentralEvent::ManufacturerDataAdvertisement {
+                                address: self.address(),
+                                manufacturer_id: k,
+                                data: v.clone(),
+                            });
                         Some((k, v.to_owned()))
                     } else {
                         warn!("Manufacturer data had wrong type: {:?}", &v.0);
@@ -309,6 +305,38 @@ impl Peripheral {
                     }
                 })
                 .collect();
+        }
+
+        if let Some(service_data) = args.service_data() {
+            properties.service_data = service_data
+                .into_iter()
+                .filter_map(|(service, data)| {
+                    let service: Uuid = service.parse().unwrap();
+                    if let Some(data) = cast::<Vec<u8>>(&data.0) {
+                        self.adapter.emit(CentralEvent::ServiceDataAdvertisement {
+                            address: self.address(),
+                            service,
+                            data: data.clone(),
+                        });
+                        Some((service, data.to_owned()))
+                    } else {
+                        warn!("Service data had wrong type: {:?}", &data.0);
+                        None
+                    }
+                })
+                .collect();
+        }
+
+        if let Some(services) = args.uuids() {
+            properties.services = services
+                .into_iter()
+                .filter_map(|uuid| uuid.parse().ok())
+                .collect();
+
+            self.adapter.emit(CentralEvent::ServicesAdvertisement {
+                address: self.address.clone(),
+                services: properties.services.clone(),
+            });
         }
 
         if let Some(address_type) = args.address_type() {
@@ -320,15 +348,19 @@ impl Peripheral {
             );
 
             properties.address_type = address_type;
+            emit_updated = true;
         }
 
         if let Some(rssi) = args.rssi() {
             let rssi = rssi as i8;
             debug!("Updating \"{}\" RSSI \"{:?}\"", self.address, rssi);
             properties.tx_power_level = Some(rssi);
+            emit_updated = true;
         }
 
-        self.adapter.emit(CentralEvent::DeviceUpdated(self.address));
+        if emit_updated {
+            self.adapter.emit(CentralEvent::DeviceUpdated(self.address));
+        }
     }
 
     pub fn proxy(&self) -> Proxy<&SyncConnection> {
--- a/src/corebluetooth/central_delegate.rs
+++ b/src/corebluetooth/central_delegate.rs
@@ -46,6 +46,9 @@ pub enum CentralDelegateEvent {
     DiscoveredPeripheral(StrongPtr),
     // Peripheral UUID, HashMap Service Uuid to StrongPtr
     DiscoveredServices(Uuid, HashMap<Uuid, StrongPtr>),
+    ManufacturerData(Uuid, u16, Vec<u8>),
+    ServiceData(Uuid, HashMap<Uuid, Vec<u8>>),
+    Services(Uuid, Vec<Uuid>),
     // DiscoveredIncludedServices(Uuid, HashMap<Uuid, StrongPtr>),
     // Peripheral UUID, HashMap Characteristic Uuid to StrongPtr
     DiscoveredCharacteristics(Uuid, HashMap<Uuid, StrongPtr>),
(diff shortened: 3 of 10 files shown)


src/api/mod.rs
```diff
@@ -377,13 +380,23 @@ pub trait Peripheral: Send + Sync + Clone + Debug {
 }
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, Copy, Clone)]
```
NZSmartie, line 380: I removed the `Copy` trait from here so that a `Vec<u8> may be supplied to the *Advertisement variants. `Vec` does not support shallow copying

src/bluez/adapter/peripheral.rs
```diff
@@ -310,6 +315,26 @@ impl Peripheral {
...
 
+        if let Some(service_data) = args.service_data() {
+            properties.service_data = service_data
+                .into_iter()
+                .filter_map(|(service, data)| {
+                    let service: UUID = service.parse().unwrap();
+                    if let Some(data) = cast::<Vec<u8>>(&data.0) {
+                        self.adapter.emit(CentralEvent::ServiceAdvertisement {
+                            address: self.address(),
+                            service: service,
```
qwandor, line 326: Nit: You can just write `service`, as the field name and variable name are the same.

src/api/mod.rs
```diff
@@ -298,6 +298,9 @@ pub struct PeripheralProperties {
     /// Advertisement data specific to the device manufacturer. The keys of this map are
     /// 'manufacturer IDs', while the values are arbitrary data.
     pub manufacturer_data: HashMap<u16, Vec<u8>>,
+    /// Advertisement data specific to the service. The keys of this map are
```
qwandor, line 301: Nit: "specific to _a_ service"

src/api/mod.rs
```diff
@@ -377,13 +380,23 @@ pub trait Peripheral: Send + Sync + Clone + Debug {
...
+    ManufacturerAdvertisement {
+        address: BDAddr,
+        manufacturer_id: u16,
+        data: Vec<u8>,
+    },
+    ServiceAdvertisement {
+        address: BDAddr,
+        service: UUID,
+        data: Vec<u8>,
+    },
```
qwandor, line 399: Having both these events and `DeviceUpdated` seems a bit inconsistent to me: now you can get some information about devices from these specific events (as well as getting a `DeviceUpdated` event), while some you only get a `DeviceUpdated` event and have to look up the peripheral.

What would you still need to use the `DeviceUpdated` event for? Just RSSI?

src/winrtble/peripheral.rs
```diff
@@ -86,11 +87,65 @@ impl Peripheral {
                     let len = reader.unconsumed_buffer_length().unwrap() as usize;
                     let mut data = vec![0u8; len];
                     reader.read_bytes(&mut data).unwrap();
+
+                    // Emit event of newly received advertisement
+                    self.adapter.emit(CentralEvent::ManufacturerAdvertisement {
+                        address: self.address,
+                        manufacturer_id: company_id,
```
qwandor, line 94: You could rename the variable to `manufacturer_id` for consistency.

src/api/mod.rs
NZSmartie: The RSSI, potentially the local name, the list of advertised services. It may also be beneficial to also include what properties have been changed in the `CentralEvent::DeviceUpdated` in for form of `bitflags`?

src/corebluetooth/peripheral.rs
```diff
@@ -76,17 +79,40 @@ impl Peripheral {
...
+                        // TODO: emit event
+                        m_clone.emit(CentralEvent::ManufacturerAdvertisement{
+                            address: properties.address,
+                            manufacturer_id,
+                            data,
+                        });
+                    },
+                    Some(CBPeripheralEvent::ServiceData(service_data)) => {
+                        let mut properties = p_clone.lock().unwrap();
+                        properties.service_data.extend(service_data.clone());
```
NZSmartie, line 105: This could be bad as the service_data doesn't get cleared out, so over time, this may use up a bit of memory. I'm not sure if the service_data reports from Core Bluetooth are complete, or one off advertisements and whether the application wants to have access to older advertisement data

NZSmartie: The next question is, in regards to the API design, do we want to retain advertisement data, or emit them and forget?

src/corebluetooth/central_delegate.rs
```diff
@@ -104,11 +106,23 @@ impl Debug for CentralDelegateEvent {
...
                 .field(uuid2)
                 .finish(),
+            CentralDelegateEvent::ManufacturerData(uuid, manufacturer_id, manufacturer_data) => f
+                .debug_tuple("ManufacturerData")
+                .field(uuid)
+                .field(manufacturer_id)
+                .field(manufacturer_data)
+                .finish(),
+            CentralDelegateEvent::ServiceData(uuid, service_data) => f
+                .debug_tuple("ManufacturerData")
```
qwandor, line 116: ServiceData

src/api/mod.rs
qwandor: Can any of these other than the RSSI actually change after a device is initially discovered?

src/winrtble/mod.rs
```diff
@@ -17,3 +17,17 @@ mod ble;
 pub mod manager;
 pub mod peripheral;
 pub mod utils;
+
+#[non_exhaustive]
+struct AdvertisementDataType;
```
qwandor, line 22: Why does this need to be a struct, rather than just a module?

src/winrtble/mod.rs
```diff
@@ -17,3 +17,17 @@ mod ble;
...
+
+// Only some of the assigned numbers are populated here.
+// Feel free to add them as needed from https://www.bluetooth.com/specifications/assigned-numbers/generic-access-profile/
+impl AdvertisementDataType {
+    ///Core Specification Supplement, Part A, section 1.11
+    pub const SERVICE_DATA_16_BIT_UUID: u8 = 0x16;
+    ///Core Specification Supplement, Part A, section 1.11
+    pub const SERVICE_DATA_32_BIT_UUID: u8 = 0x20;
+    ///Core Specification Supplement, Part A, section 1.11
```
qwandor, line 31: Nit: Having the same doc comment on all three constants is a bit repetitive, maybe it belongs on the module instead?

src/winrtble/peripheral.rs
```diff
@@ -81,13 +86,66 @@ impl Peripheral {
...
+                        manufacturer_id,
+                        data: data.clone(),
+                    });
+
+                    (manufacturer_id, data)
+                })
+                .collect();
+        }
+
+        if let Ok(data_sections) = advertisement.data_sections() {
```
qwandor, line 108: It's probably worth adding a comment here that we have to do this because the Windows API for some reason doesn't parse service data for us.

src/winrtble/peripheral.rs
```diff
@@ -81,13 +86,66 @@ impl Peripheral {
...
+            properties.service_data = data_sections
+                .into_iter()
+                .filter_map(|d| {
+                    let buffer = d.data().unwrap();
+                    let reader = DataReader::from_buffer(&buffer).unwrap();
+                    let len = reader.unconsumed_buffer_length().unwrap() as usize;
+                    let mut data = vec![0u8; len];
+                    reader.read_bytes(&mut data).unwrap();
```
qwandor, line 116: It might be worth factoring out a helper function to convert a buffer to a `Vec<u8>`, as there are at least two places doing the same thing.

src/winrtble/peripheral.rs
```diff
@@ -81,13 +86,66 @@ impl Peripheral {
...
+                            let (uuid, data) = data.split_at(4);
+                            let uuid = uuid_from_u32(u32::from_le_bytes(uuid.try_into().unwrap()));
+                            Some((uuid, data.to_owned()))
+                        }
+                        AdvertisementDataType::SERVICE_DATA_128_BIT_UUID => {
+                            let (uuid, data) = data.split_at(16);
+                            let uuid = Uuid::from_slice(uuid).unwrap();
+                            Some((uuid, data.to_owned()))
+                        }
+                        _ => None,
```
qwandor, line 134: If you `return None` here, then you can avoid the if let immediately after.

src/corebluetooth/central_delegate.rs
```diff
@@ -340,6 +354,51 @@ pub mod CentralDelegate {
             delegate,
             CentralDelegateEvent::DiscoveredPeripheral(held_peripheral),
         );
+
+        let puuid_nsstring = ns::uuid_uuidstring(cb::peer_identifier(peripheral));
+        let puuid = Uuid::from_str(&NSStringUtils::string_to_string(puuid_nsstring)).unwrap();
+
+        let manufacturer_data = ns::dictionary_objectforkey(_adv_data, unsafe {
```
qwandor, line 361: Now that you are using `_adv_data`, remote the underscore at the start. It was just there to indicate that the parameter was unused.

src/winrtble/mod.rs
NZSmartie: I don't remember, and looking back it it now, it makes more sense to be a module

src/winrtble/mod.rs
NZSmartie: it was copy/pasted directly from the assigned numbers. I may just leave a module level comment saying to find out more information there

src/corebluetooth/central_delegate.rs
```diff
@@ -340,6 +354,51 @@ pub mod CentralDelegate {
...
+            cb::ADVERTISEMENT_DATA_MANUFACTURER_DATA_KEY
+        });
+        if manufacturer_data != nil {
+            // manufacturer_data: NSData
+            let length = ns::data_length(manufacturer_data);
+            if length >= 2 {
+                let bytes = ns::data_bytes(manufacturer_data);
+                let v = unsafe { slice::from_raw_parts(bytes, length as usize).to_vec() };
```
qwandor, line 369: Seeing as all you are doing with v is splitting it into two slices which you then copy, there's no need for the `to_vec()` here.

src/api/mod.rs
NZSmartie: I see that `tx_power_level` may also change since it's part of the Assigned Number for Advertisement Data https://www.bluetooth.com/specifications/assigned-numbers/generic-access-profile/

src/winrtble/peripheral.rs
NZSmartie: That was a good idea, Thank you!

src/winrtble/peripheral.rs
NZSmartie: I've since created `utils::to_vec()`

src/winrtble/peripheral.rs
NZSmartie: I've added a comment, I hope it explains why, well enough

src/corebluetooth/peripheral.rs
qwandor: I think it's fine as the service UUIDs will be the same, or at least a small set, so they will be replaced in the HashMap won't they?

src/corebluetooth/peripheral.rs
```diff
@@ -76,17 +79,42 @@ impl Peripheral {
...
+                                value: data,
+                            },
+                        );
+                    }
+                    Some(CBPeripheralEvent::ManufacturerData(manufacturer_id, data)) => {
+                        let mut properties = p_clone.lock().unwrap();
+                        properties
+                            .manufacturer_data
+                            .insert(manufacturer_id, data.clone());
+                        // TODO: emit event
```
qwandor, line 98: You're already doing that, aren't you?

src/winrtble/peripheral.rs
qwandor: Looks good to me.

src/winrtble/peripheral.rs
```diff
@@ -101,6 +150,8 @@ impl Peripheral {
             .raw_signal_strength_in_dbm()
             .ok()
             .map(|rssi| rssi as i8);
+
+        self.adapter.emit(CentralEvent::DeviceUpdated(self.address));
```
qwandor, line 154: It looks like you send this even if the only change is the service data or manufacturer data, which you already send an event for above, whereas in the Mac OS implementation I think you only send the `ServiceAdvertisement`/`ManufacturerAdvertisement` event.

I suggest that you only send the `DeviceUpdated` event if some other property of the device has changed.

src/bluez/adapter/peripheral.rs
```diff
@@ -311,6 +316,26 @@ impl Peripheral {
...
+                        });
+                        Some((service, data.to_owned()))
+                    } else {
+                        warn!("Service data had wrong type: {:?}", &data.0);
+                        None
+                    }
+                })
+                .collect();
+        }
+
```
qwandor, line 338: Ditto here, let's only send the `DeviceUpdated` event if some property other than service/manufacturer data changed.

src/corebluetooth/peripheral.rs
NZSmartie: According to [`impl<K, V, S> Extend<(K, V)> for HashMap<K, V, S>`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-Extend%3C(K%2C%20V)%3E)
> Inserts all new key-values from the iterator and replaces values with existing keys with new values returned from the iterator.

qwandor: Right. So if the UUIDs are the same, which they generally will be, then the map won't get any bigger.

src/corebluetooth/peripheral.rs
```diff
@@ -55,39 +55,76 @@ impl Peripheral {
...
         let notification_handlers = Arc::new(Mutex::new(Vec::<NotificationHandler>::new()));
         let nh_clone = notification_handlers.clone();
+        let p_clone = properties.clone();
+        let m_clone = manager.clone();
         task::spawn(async move {
             let mut event_receiver = event_receiver;
             loop {
-                let event = event_receiver.next().await;
-                if event.is_none() {
+                if let Some(event) = event_receiver.next().await {
```
NZSmartie, line 78: This match statement was changed from an `Option<T>` to just `T`. so that each variant can be caught at compile time and not only at runtime only when someone is paying attention to it.

src/corebluetooth/peripheral.rs
```diff
@@ -55,39 +55,76 @@ impl Peripheral {
...
+                    match event {
```
NZSmartie, line 79: This match statement was changed from an Option<T> to just T. so that each variant can be caught at compile time and not only at runtime only when someone is paying attention to it.

src/corebluetooth/central_delegate.rs
```diff
@@ -340,6 +360,66 @@ pub mod CentralDelegate {
...
+            }
+
+            send_delegate_event(delegate, CentralDelegateEvent::ServiceData(puuid, result));
+        }
+
+        let services = ns::dictionary_objectforkey(adv_data, unsafe {
+            cb::ADVERTISEMENT_DATA_SERVICE_UUIDS_KEY
+        });
+        if services != nil {
+            // service_data: [CBUUID, NSData]
```
qwandor, line 413: This comment looks wrong

src/corebluetooth/peripheral.rs
qwandor: Thanks, I agree this is better. (-:

qwandor: You could, optionally, combine the `if let` and `match` into a single match, like:

```rust
match event_receiver.next().await {
  Some(...) => {...}
  Some(...) => {...}
  None => {error!(...); break;}
}
```

You'd get the same enforcement that all enum variants are covered, with a wee bit less code.

qwandor (approved):
A couple of minor comments, but looks good once those are resolved.

src/corebluetooth/central_delegate.rs
qwandor: Fixed myself.

src/corebluetooth/peripheral.rs
qwandor: Done myself.

PR #127, opened 2021-02-22
Update from winrt to windows crate.

qwandor:


NZSmartie (approved):
Looks good to me, examples worked on my machine, though I haven't tested it with an existing project yet

src/winrtble/utils.rs
```diff
@@ -21,17 +21,19 @@ use bindings::windows::devices::bluetooth::generic_attribute_profile::{
...
 pub fn to_error(status: GattCommunicationStatus) -> Result<()> {
-    match status {
-        GattCommunicationStatus::AccessDenied => Err(Error::PermissionDenied),
-        GattCommunicationStatus::Unreachable => Err(Error::NotConnected),
-        GattCommunicationStatus::Success => Ok(()),
-        GattCommunicationStatus::ProtocolError => {
-            Err(Error::NotSupported("ProtocolError".to_string()))
-        }
-        _ => Err(Error::Other(format!("Communication Error:"))),
+    if status == GattCommunicationStatus::AccessDenied {
```
qdot, line 27: Oof. Gonna miss this being an enum. :(

qdot (approved):

src/winrtble/utils.rs
qwandor: Was it actually an enum before, or did it just derive `PartialEq` and `Eq` rather than implementing them manually? Apparently that makes a difference on whether a constant is allowed to be used in a match expression: https://rust-lang.github.io/rfcs/1445-restrict-constants-in-patterns.html

qdot: > Was it actually an enum before, or did it just derive `PartialEq` and `Eq` rather than implementing them manually? Apparently that makes a difference on whether a constant is allowed to be used in a match expression: https://rust-lang.github.io/rfcs/1445-restrict-constants-in-patterns.html

Oh huh, I had no idea! So you're right, implementation may've just been different. Would this be worth filing on the windows project for, to try and get pattern matching back on these (and other) types?

qwandor: I have no strong opinion either way. It sounds like the compiler behaviour will change in future though, if I'm reading that RFC correctly? So maybe it's better just to wait for that.

PR #128, opened 2021-02-26
Fix serde support for Uuid type

NZSmartie:
To maintain a `serde` feature with the btleplug package, some package
renaming magic needed to happen. See
- https://blog.turbo.fish/cargo-features/
- https://github.com/serde-rs/serde/issues/1990

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,15 @@ name = "btleplug"
 path = "src/lib.rs"
 # crate-type = ["staticlib"]
 
+[features]
+serde = ["uuid/serde", "serde_cr"]
+
 [dependencies]
 log = "0.4.14"
 bitflags = "1.2.1"
 thiserror = "1.0.23"
 uuid = "0.8.2"
-serde = { version = "1.0.123", features = ["derive"], default-features = false, optional = true }
+serde_cr = { package = "serde", version = "1.0.123", features = ["derive"], default-features = false, optional = true }
 dashmap = "4.0.2"
 futures = "0.3.12"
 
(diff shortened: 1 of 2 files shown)


NZSmartie:
It may also be a good idea to add `--all-features` to `cargo check` in the CI

Cargo.toml
```diff
@@ -20,12 +20,15 @@ name = "btleplug"
...
+[features]
+serde = ["uuid/serde", "serde_cr", "serde_cr/derive"]
+
 [dependencies]
 log = "0.4.14"
 bitflags = "1.2.1"
 thiserror = "1.0.23"
 uuid = "0.8.2"
-serde = { version = "1.0.123", features = ["derive"], default-features = false, optional = true }
+serde_cr = { package ="serde", version = "1.0.123", features = ["derive"], optional = true }
```
qwandor, line 31: Nit: you're missing a space after one of your `=`.

Cargo.toml
```diff
@@ -20,12 +20,15 @@ name = "btleplug"
 path = "src/lib.rs"
 # crate-type = ["staticlib"]
 
+[features]
+serde = ["uuid/serde", "serde_cr", "serde_cr/derive"]
```
qwandor, line 24: Huh, why does the 'derive' feature need to be specified both here and below?

Cargo.toml
qwandor, line 31: Can you keep `default-features = false` here?

Cargo.toml
qwandor: I just checked and it looks like you don't need it here.

src/api/mod.rs
```diff
@@ -18,6 +18,8 @@ use crate::{Error, Result};
 pub use adapter_manager::AdapterManager;
 use bitflags::bitflags;
 #[cfg(feature = "serde")]
+extern crate serde_cr as serde;
```
qwandor, line 21: Nit: I think `use` is more typical in 2018 edition Rust.

qwandor (approved):

PR #130, opened 2021-02-28
Log messages de-escalation

qdot:
Does what it says in the title. :3

### DIFF
--- a/src/corebluetooth/central_delegate.rs
+++ b/src/corebluetooth/central_delegate.rs
@@ -23,7 +23,7 @@ use super::{
 use futures::channel::mpsc::{self, Receiver, Sender};
 use futures::sink::SinkExt;
 use libc::{c_char, c_void};
-use log::{error, info, trace};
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -23,7 +23,7 @@ path = "src/lib.rs"
 [dependencies]
 log = "0.4.14"
 bitflags = "1.2.1"
-thiserror = "1.0.23"
+thiserror = "1.0"
```
qwandor, line 26: A newer version should be fine, but I'm wary of building against an older version than we expect. Why do you need to change this dependency?

qdot: On both of these, I was seeing collisions between btleplug and other libraries I write farther up the stack. I suppose I could specify versions at the top level in that case? Is there a standard for this in the community?

qwandor (approved):
Looks good once you rebase and update the title.

NZSmartie (approved):

PR #135, opened 2021-03-07
Reintroduce `cargo check --all-features` and `cargo check --no-default-features`

NZSmartie:
the crate _serde_bytes_ is also added as it's needed for `Vec<u8>` and `Vec<Uuid>`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ path = "src/lib.rs"
 # crate-type = ["staticlib"]
 
 [features]
-serde = ["uuid/serde", "serde_cr"]
+serde = ["uuid/serde", "serde_cr", "serde_bytes"]
 
 [dependencies]
 log = "0.4.14"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,14 +21,15 @@ path = "src/lib.rs"
...
-serde = ["uuid/serde", "serde_cr"]
+serde = ["uuid/serde", "serde_cr", "serde_bytes"]
 
 [dependencies]
 log = "0.4.14"
 bitflags = "1.2.1"
 thiserror = "1.0.23"
 uuid = "0.8.2"
 serde_cr = { package = "serde", version = "1.0.123", features = ["derive"], default-features = false, optional = true }
+serde_bytes = {version = "0.11.5", optional=true }
```
qwandor, line 32: Nit: spaces around the `=`, to match the rest of the file.

qwandor (approved):
I don't understand how this works, but apparently it does.

PR #139, opened 2021-03-19
Support handles starting alphanumerically

tsjost:
With a handle like `/org/bluez/hci0/dev_01_02_03_04_05_06/serviceff00`, `trim_start_matches(char::is_alphabetic)` will strip off `serviceff` and then `[..4]` on `00` will run out of bounds and crash/panic.


qwandor:
Thanks for the PR!

I believe this is fixed in the dev branch along with a bunch of other changes (such as moving to async). Can you see if you still have any problems on that?

PR #144, opened 2021-04-05
Add custom de-/serialization for BDAddr

MattesWhite:
Like suggested in #68 this PR adds options to de-/serialize `BDAddr` in different formats.

If this get merged I could do a similar solution for the UUIDs.

### DIFF
--- a/src/api/bdaddr.rs
+++ b/src/api/bdaddr.rs
@@ -4,17 +4,7 @@ use std::convert::{TryFrom, TryInto};
 use std::fmt::{self, Debug, Display, Formatter, LowerHex, UpperHex};
 use std::str::FromStr;
 
-#[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
-#[cfg(feature = "serde")]
-use serde_cr as serde;
-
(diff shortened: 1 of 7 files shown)


src/api/bdaddr.rs
```diff
@@ -203,6 +193,163 @@ impl BDAddr {
...
+    /// ```
+    /// //use btleplug::api::BDAddr;
+    /// //use serde::{Serialize, Deserialize};
+    /// //^^^ Error can't find crate `serde`???
+    /// //# use serde_cr as serde;
+    /// //
+    /// //#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+    /// //struct S {
+    /// //    addr: BDAddr,
+    /// //}
```
MattesWhite, line 242: I would like to write further examples in this style. However, when I try to compile this I get an error _crate not found_. even with the `use serde_cr as serde;`. Has anyone an idea how to fix this? (Yes, I did use `--all-features` to enable the `serde` feature.

MattesWhite: Okay got it myself

qwandor (approved):

PR #151, opened 2021-04-24
Added subscribe on notify characteristic example

blandger:
I wonder why there are 5 changed files instead of 1 added examples ?
I have all are in sync with origin/dev.
Example doesn't do reconnect after loosing BLE. I think now to implement is properly...


examples/subscribe_notify_characteristic.rs
```diff
@@ -0,0 +1,118 @@
...
+
+const PERIPHERAL_NAME_MATCH_FILTER: &'static str = "Neuro"; // filter BLE device by partial name
+const NOTIFY_CHARACTERISTIC_MATCH_UUID: &'static str = "6E400002-B534-F393-67A9-E50E24DCCA9E"; // subscribe UUID
+
+#[cfg(target_os = "linux")]
+fn print_adapter_info(adapter: &Adapter) {
+    println!("connected adapter {:?}", adapter);
+}
+
+#[cfg(any(target_os = "windows", target_os = "macos"))]
```
qwandor, line 20: This platform-specific stuff shouldn't be necessary anymore.

examples/subscribe_notify_characteristic.rs
```diff
@@ -0,0 +1,118 @@
...
+use btleplug::platform::{Adapter, Manager};
+use btleplug::api::{Central, Manager as _, Peripheral};
+use std::io::Cursor;
+use std::error::Error;
+use uuid::Uuid;
+use tokio::time;
+use futures::stream::{StreamExt};
+
+const PERIPHERAL_NAME_MATCH_FILTER: &'static str = "Neuro"; // filter BLE device by partial name
+const NOTIFY_CHARACTERISTIC_MATCH_UUID: &'static str = "6E400002-B534-F393-67A9-E50E24DCCA9E"; // subscribe UUID
```
qwandor, line 13: It should be possible to make the constants `Uuid`s rather than strings.

qwandor:
Can you rebase on the latest dev branch with just the commits you actually want to include? This PR currently has a bunch of other unrelated changes.

examples/subscribe_notify_characteristic.rs
blandger: ok, I will remove it

examples/subscribe_notify_characteristic.rs
blandger: That struct + method should I use. I'd like to... my previous code used const func to parse uuid value which I don't see in dev now.

blandger:
> Can you rebase on the latest dev branch with just the commits you actually want to include? This PR currently has a bunch of other unrelated changes.

I will try.
The strange what I have local dev branch synced with remote origin dev but there are other changes show in my PR.

examples/subscribe_notify_characteristic.rs
qwandor: You can do `Uuid::from_u128(0x6e400002_b534_f393_67a9_e50e24dccA9e)`.

PR #154, opened 2021-04-30
BlueZ: Set discovery filter transport to auto

NZSmartie:
Set the Transport filter to "auto" when starting a scan, to interleave both BR/EDR and LE devices. In some cases (myself affected) only BR/EDR devices were being discovered when this filter was not set.

Tested with Ubuntu 20.04, BlueZ 5.53


qwandor:
It is surprising that this makes a difference, as `auto` is supposed to be the default. I wonder if the issue is actually with whether there is a filter at all or no filter. Can you test (on the same machine and with the same devices) whether not setting this, but setting some unrelated filter attribute to it's default value (e.g. set `discoverable` to `false`) has the same effect?

qdot:
Bringing this in as a cherry-pick to dev, and filing #154 as the followup.

PR #156, opened 2021-05-02
Force  to read from GattCharacteristic from device instead of cache

### LINKED ISSUE #155: Windows reads Gatt Characteristic from Windows Cache by default
**Describe the bug**
I have a bluetoothLE temperature sensor. When reading a Gatt Characteristic from windows, the first value read after a sensor connect is always the same value until I disonnect / reconnect the device.

**Expected behavior**
Running the same program under Linux I get changing temperature values with each read, without needing to reconnect the device everytime.

**Actual behavior**
See above.

**Additional context**
I think windows per default always reads from the Windows Cache if it is available. Direct quote from the [documentation](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattcharacteristic.readvalueasync?view=winrt-19041):

`ReadValueAsync():     Performs a Characteristic Value read from the value cache maintained by Windows.`

I do have a fix (see the [diff here](https://github.com/deviceplug/btleplug/compare/dev...jsosulski:win_force_read_from_device?expand=1).) that solves the problem but unfortunately I am not aware of any side effects and whether you would want to expose the option whether to read from cache or not. However, this would unify behaviour between linux and windows for now.

jsosulski:
Closes #155 

For some reason `BluetoothCacheMode(1)` is the no caching mode and `BluetoothCacheMode(0)` uses the cache. See [docs](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothcachemode?view=winrt-19041#fields)

Using `BluetoothCacheMode::Uncached` did not work for me. Maybe this is not really an enum? My IDE was also not very helpful regarding these windows bindings.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -17,6 +17,7 @@ fn main() {
         Windows::Devices::Bluetooth::{
             BluetoothConnectionStatus,
             BluetoothLEDevice,
+            BluetoothCacheMode,
         },
         Windows::Devices::Radios::{
             Radio,
--- a/src/winrtble/ble/characteristic.rs
+++ b/src/winrtble/ble/characteristic.rs
@@ -68,7 +68,13 @@ impl BLECharacteristic {
     }
 
     pub async fn read_value(&self) -> Result<Vec<u8>> {
-        let result = self.characteristic.ReadValueWithCacheModeAsync(BluetoothCacheMode(1)).unwrap().await.unwrap();
+        let result = self
+            .characteristic
+            // 1 = Ignore cache, 0 = Use windows cache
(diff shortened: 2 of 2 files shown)


jsosulski:
Found the issue why I couldnt access it like an enum. It apparently is a [tuple struct](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Devices/Bluetooth/struct.BluetoothCacheMode.html).

qwandor (commented):
Looks good to me, but Kyle should probably take a look for Windows.

qdot (approved):
lgtm

PR #180, opened 2021-08-14
winrtble: report characteristic properties

rib:
While experimenting with using btleplug on Windows with heart rate monitors where I was expecting to see characteristics with the notify property I noticed that no characteristics had any properties.

Poking around I saw that the `utils.rs::to_char_props()` utility hadn't been implemented yet and it was just returning blank flags.

After implementing this I can now see properties as expected.

As a side note: I noticed on Windows there are two additional properties documented here: https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattcharacteristicproperties?view=winrt-20348

`ReliableWrites` and `WritableAuxiliaries`

I wonder if `CharPropFlags` should be expanded to a u32 as on Windows and these properties could be reported too?

### DIFF
--- a/src/winrtble/utils.rs
+++ b/src/winrtble/utils.rs
@@ -56,8 +56,39 @@ pub fn to_guid(uuid: &Uuid) -> Guid {
     Guid::from_values(data1, data2, data3, data4.to_owned())
 }
 
-pub fn to_char_props(_: &GattCharacteristicProperties) -> CharPropFlags {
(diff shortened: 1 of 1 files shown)


qwandor:
Thanks for the PR! Please rebase against the dev branch and run `cargo fmt`.

qwandor (approved):

PR #184, opened 2021-08-15
Winrtble: aggregate share Peripheral state under one boxed Arc

rib:
_Note: this is based on https://github.com/deviceplug/btleplug/pull/183, so the first patch can be ignored here_

When I was looking at making changes in the winrtble backend I noticed that there were lots of separately boxed internal struct members and as far as I could see the only purpose for putting this state under `Arc<>`s was because a `Peripheral` is expected to implement Clone and so conceptually all this internal state can effectively be shared in case a Peripheral gets `clone()`d multiple times.

As is; each `Arc<>` is a separate heap allocation with a seperate ref count though and the derived implementation of `Clone` will necessarily have to increment each separate ref count when they could instead all be grouped together into a single heap allocation and `Clone` can be reduced to a single reference increment.

Since it looks like there are other structs in btleplug that also similarly have multiple individial `Arc` wrappers for members it could perhaps make sense to consider similar changes elsewhere but to start with I just wanted to see if was straight-forward to affect the change in one place and get feedback on whether it seems like a reasonable change.

### DIFF
--- a/src/winrtble/peripheral.rs
+++ b/src/winrtble/peripheral.rs
@@ -44,39 +44,40 @@ use bindings::Windows::Devices::Bluetooth::Advertisement::*;
 /// Implementation of [api::Peripheral](crate::api::Peripheral).
 #[derive(Clone)]
 pub struct Peripheral {
-    device: Arc<tokio::sync::Mutex<Option<BLEDevice>>>,
-    adapter: AdapterManager<Self>,
+    shared: Arc<Shared>,
+}
+
+struct Shared {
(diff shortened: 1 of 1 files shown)


qdot:
Awesome, I'll try to take a look at this soon.

Also, if you're seeing things that look like they could pop up elsewhere, feel free to file followup issues just to make sure they don't get lost. I am *super* flakey and sometimes totally miss that after landing and closing things out, heh.

qwandor (approved):

qdot:
@rib Looks like this might need a manual rebase since I brought in #183 first? Once that's handled this should be landable.

Issue #182, opened 2021-08-15
Discussion: winrt Peripheral state lifetime/ownership design details


rib:
Sorry this is more of a question/discussion than an issue...

While looking to use btleplug on Windows with heart rate monitors I wasn't exactly sure if I was responsible for cleaning up anything related to a peripheral if I received a disconnect event - in particular I wasn't sure if I would need to call `notifications()` again to get a stream for events if I were to reconnect and I wasn't sure if I would need to explicitly re-subscribe to characteristic notifications if I reconnect.

Looking at the code for windows and also comparing a little with bluez I think maybe there could be some inconsistency here at the moment, and in relation to this I saw that there's various peripheral state held under separate Arc<>s that made me wonder if there's a design requirement necessitating this that wasn't obvious to me.

With the `notifications()` stream; I see in the winrt backend that each time this would be called then a new `mpsc` channel will be created and this will be added to an internal vector held within an Arc<Mutex<>>:
```rust
    async fn notifications(&self) -> Result<Pin<Box<dyn Stream<Item = ValueNotification> + Send>>> {
        let (sender, receiver) = mpsc::unbounded();
        let mut senders = self.notification_senders.lock().unwrap();
        senders.push(sender);
        Ok(Box::pin(receiver))
    }
```

There's nothing that I've seen that would ever remove from that vector and while my understanding is that Peripherals themselves will persist between device disconnect/connect cycles then it seems notable that if an application were to repeatedly call `notifications()` they would end up with an ever-growing vector of channels.

By comparison the bluez backend delegates to a very similar api from the bluez library which return a filtered stream for events and doesn't seem to claim any kind of reference to that stream internally - it just passes it on to the application. Looking at the bluez implementation of that API it also looks like it's been carefully designed to hand out a stream that filters a central dbus connection and has a `Drop` implementation that will asynchronously make sure to clean up any associated internal state. So I think with the bluez backend an application should be able to call `notifications()` multiple times and so long as it drops previously received streams then there wouldn't be any kind of growing collection of internal state.

I was wondering if it could instead make sense for the winrt backend to instead just create a single `sync::broadcast` channel and clone receivers from this to hand out as a stream and thereby avoid needing to maintain an internal collection of channels/receivers? (A `broadcast::Receiver` implements Drop which will handle removing itself from the channel)

When considering this I also noticed that the current Peripheral imlementation is tracking several pieces of state under Arc<Mutex<>> wrappers:
```rust
pub struct Peripheral {
[...]

PR #185, opened 2021-08-15
Central broadcast channel events

rib:
_Note: this PR is based on top of https://github.com/deviceplug/btleplug/pull/184 so the first two patches can be ignored_

As another follow up to #182 this makes equivalent changes to the `AdapterManager` to use a broadcast channel for streaming `CentralEvent`s; avoiding the need to track a set of `mpcs` channels and the need to clean up stale senders.

### DIFF
--- a/src/common/adapter_manager.rs
+++ b/src/common/adapter_manager.rs
@@ -11,31 +11,37 @@
 // following copyright:
 //
 // Copyright (c) 2014 The Rust Project Developers
-use crate::{
-    api::{BDAddr, CentralEvent, Peripheral},
-    common::util::send_notification,
-};
+use crate::api::{BDAddr, CentralEvent, Peripheral};
 use dashmap::{mapref::one::RefMut, DashMap};
-use futures::channel::mpsc::{self, UnboundedSender};
-use futures::stream::Stream;
+use futures::stream::{Stream, StreamExt};
+use log::trace;
 use std::pin::Pin;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
+use tokio::sync::broadcast;
+use tokio_stream::wrappers::BroadcastStream;
 
 #[derive(Clone, Debug)]
 pub struct AdapterManager<PeripheralType>
 where
     PeripheralType: Peripheral,
 {
-    peripherals: Arc<DashMap<BDAddr, PeripheralType>>,
-    async_senders: Arc<Mutex<Vec<UnboundedSender<CentralEvent>>>>,
+    shared: Arc<Shared<PeripheralType>>,
+}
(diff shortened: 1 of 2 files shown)


src/common/adapter_manager.rs
```diff
@@ -47,50 +52,59 @@ where
...
-                self.peripherals.remove(&addr);
+                self.shared.peripherals.remove(&addr);
             }
             _ => {}
         }
 
-        send_notification(&self.async_senders, &event);
+        // Note: we ignore possible send errors which are expected while there
+        // could be zero recievers if application is not subscribed
+        let _ = self.shared.events_channel.send(event);
```
qwandor, line 65: Rather than ignoring the error entirely, how about logging it at trace level?

src/common/adapter_manager.rs
```diff
@@ -47,50 +52,59 @@ where
...
-        self.async_senders.lock().unwrap().push(sender);
-        Box::pin(receiver)
+        let receiver = self.shared.events_channel.subscribe();
+        Box::pin(BroadcastStream::new(receiver).filter_map(|x| async move {
+            if x.is_ok() {
+                Some(x.unwrap())
+            } else {
+                None
+            }
+        }))
```
qwandor, line 76: You have the same pattern somewhere else in a previous PR, could you factor it out to a helper function?

src/common/adapter_manager.rs
rib: okey, yip, I've pushed an update which traces any lost events due to having no subscribed receivers

src/common/adapter_manager.rs
rib: Although it's equivalent to what's done for the characteristic notification events I wonder if it's worth having a utility while it's semantically a different event stream (theoretically it seems plausible that it could make sense to do different things here than is done for a characteristic notification stream)?

There's a repetition of the same thing between the winrt backend and corebluetooth backend too and I suppose a utility could be added for those, which could probably technically be used here too, but maybe it'd be more worthwhile having a wrapper if it's also for a more specific purpose like for a notification stream?

Maybe I could add a utility at the same time as updating the corebluetooth backend - just for notifications to start with and also remove `send_notification` which is now not used by anything?

rib: Okey, for reference here I've updated my PR that updates the corebluetooth backend to add a utility for this: https://github.com/deviceplug/btleplug/pull/187

https://github.com/deviceplug/btleplug/pull/187/commits/76c0fc39c1ccfacd4b4f54bc1033e5f3402ba9fa

qwandor (approved):

PR #186, opened 2021-08-15
Corebluetooth broadcast channel notifications

rib:
_Note: this PR is based on https://github.com/deviceplug/btleplug/pull/185 so the first three patches can be ignored_

This (blindly) makes the same change as made in https://github.com/deviceplug/btleplug/pull/183 and https://github.com/deviceplug/btleplug/pull/184 for the corebluetooth backend. (untested)


rib:
(sorry I meant to open this against the dev branch)

PR #188, opened 2021-08-20
Wrote `is_connected` for CoreBluetooth

MacTuitui:
I tried to keep things minimal for my first PR.
I had to expose a new `CBPeripheralState` to get the correct state. 
Let me know if that works for you.

### DIFF
--- a/src/corebluetooth/framework.rs
+++ b/src/corebluetooth/framework.rs
@@ -289,7 +289,6 @@ pub mod cb {
         unsafe { msg_send![cbperipheral, name] }
     }
 
-    //returns
     #[derive(Copy, Clone, Debug, Eq, PartialEq)]
     #[repr(i64)]
     pub enum CBPeripheralState {
@@ -298,6 +297,7 @@ pub mod cb {
         Connected = 2,
         Disconnecting = 3,
     }
+
     pub fn peripheral_state(cbperipheral: *mut Object) -> CBPeripheralState {
         unsafe { msg_send![cbperipheral, state] }
     }
--- a/src/corebluetooth/framework.rs
+++ b/src/corebluetooth/framework.rs
@@ -289,7 +289,16 @@ pub mod cb {
         unsafe { msg_send![cbperipheral, name] }
     }
 
-    pub fn peripheral_state(cbperipheral: *mut Object) -> c_int {
+    //returns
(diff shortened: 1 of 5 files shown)


src/corebluetooth/framework.rs
```diff
@@ -289,7 +289,16 @@ pub mod cb {
         unsafe { msg_send![cbperipheral, name] }
     }
 
-    pub fn peripheral_state(cbperipheral: *mut Object) -> c_int {
+    //returns
```
qwandor, line 292: Remove this comment, I'm not sure what it's doing here.

src/corebluetooth/framework.rs
```diff
@@ -289,7 +289,16 @@ pub mod cb {
...
+    #[derive(Copy, Clone, Debug, Eq, PartialEq)]
+    #[repr(i64)]
+    pub enum CBPeripheralState {
+        Disonnected = 0,
+        Connecting = 1,
+        Connected = 2,
+        Disconnecting = 3,
+    }
```
qwandor, line 300: Add a blank line between here and the function below.

src/corebluetooth/internal.rs
```diff
@@ -481,6 +484,15 @@ impl CoreBluetoothInternal {
             cb::centralmanager_connectperipheral(*self.manager, *p.peripheral);
         }
     }
```
qwandor, line 486: Add a blank line here before the new method.

src/corebluetooth/peripheral.rs
```diff
@@ -119,6 +122,7 @@ impl Peripheral {
                     }
                 }
             }
+            //drop?
```
qwandor, line 125: What does this comment mean?

src/corebluetooth/peripheral.rs
```diff
@@ -177,7 +181,23 @@ impl api::Peripheral for Peripheral {
 
     async fn is_connected(&self) -> Result<bool> {
         // TODO
```
qwandor, line 183: You can remove the TODO now.

qwandor:
This whole thing is a mess and needs some refactoring to simplify it, but your change looks generally good. I've left some comments with style nits to fix, and please run `cargo fmt`, then we should be able to merge this.

qwandor (approved):

PR #198, opened 2021-08-25
Expose RSSI property

rib:
This adds rssi to PeripheralProperties to complement tx_power_level
that's already reported.

This exposes the rssi as i16 which is consistent with bluez and Windows.

This also updates tx_power_level to be an i16 for consistency - though
it is a small public API change.

Note: the corebluetooth backend currently always reports None for the TX
and RSSI power levels.

Addresses: #76


qwandor:
Thanks! Looks good, but can you add a line to CHANGELOG.md for this as a breaking change?

qwandor (approved):

PR #199, opened 2021-08-25
Removes 'discovery_count' debug value

rib:
This discovery_count debug value was only maintained in the winrt
backend (it was a count of advertising packets received)

The documentation for the value didn't really reflect what the value was
and since it's not updated in the corebluetooth or bluez backends it's
not really something that can be depended on by applications.


qwandor:
Seems reasonable to me. @qdot was there any reason for having this?

PR #208, opened 2021-09-03
Add ScanFilter to Central::start_scan()

paroga:
The [`requestDevice`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetooth-requestdevice) method from _WebBluetooth_ requires a list of service uuids to allow access to characteristics. See [the code](https://github.com/deviceplug/btleplug/blob/2cb6cb233dbcacfe9b79ef9ba1289b66ede0e8cb/src/wasm/adapter.rs#L99-L101) of #204 for more details.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -180,6 +180,14 @@ pub struct PeripheralProperties {
     pub services: Vec<Uuid>,
 }
 
+/// The filter used when scanning for BLE devices.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct ScanFilter {
+    /// If the filter contains at least one service UUID, only devices supporting at least one of
+    /// the given services will be available.
+    pub services: Vec<Uuid>,
+}
+
 /// The type of write operation to use.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum WriteType {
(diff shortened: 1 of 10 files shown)


src/api/mod.rs
```diff
@@ -180,6 +180,14 @@ pub struct PeripheralProperties {
     pub services: Vec<Uuid>,
 }
 
+/// The filter used when scanning for BLE devices.
+#[derive(Debug, Default, Clone)]
+pub struct ScanFilter {
```
qwandor, line 185: Derive `Eq` and `PartialEq` too, they can be useful for tests.

src/api/mod.rs
```diff
@@ -299,7 +307,11 @@ pub trait Central: Send + Sync + Clone {
     /// Starts a scan for BLE devices. This scan will generally continue until explicitly stopped,
     /// although this may depend on your Bluetooth adapter. Discovered devices will be announced
     /// to subscribers of `events` and will be available via `peripherals()`.
-    async fn start_scan(&self) -> Result<()>;
+    /// The filter can be used to scan only for specific devices. While some implementations migth
```
qwandor, line 310: Typo: might

qwandor (approved):

qwandor:
Are you able to implement this on MacOS and Windows too?

paroga:
Most of the work for the macOS port is already done and there are only a [few lines of code missing](https://github.com/deviceplug/btleplug/blob/f7a440a85a3bf38153a595ac0462de95c407cea9/src/corebluetooth/internal.rs#L806). Only a `NSArray` with the Uuids [needs to be set](https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerscanoptionsolicitedserviceuuidskey), but I don't have any ObjC experience, to know how to handle the ownership correctly. I could try to get it compile, but unfortunately I don't have a physical machine to test it on.

qwandor:
If you can send another PR for the MacOS implementation that at least compiles and looks correct, I can test it on my machine.

PR #209, opened 2021-09-07
Implemented is_connected() for OSX

coral:
Hello!

Currently quite hard to actually track the device state when using CoreBluetooth so I implemented is_connected() against the objc state function and resolved it into an new enum called CoreBluetoothState.

### DIFF
--- a/src/corebluetooth/internal.rs
+++ b/src/corebluetooth/internal.rs
@@ -110,8 +110,17 @@ impl CBCharacteristic {
 pub enum CoreBluetoothReply {
     ReadResult(Vec<u8>),
     Connected(BTreeSet<Characteristic>),
+    State(CoreBluetoothState),
     Ok,
     Err(String),
+}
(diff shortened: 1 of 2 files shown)


src/corebluetooth/internal.rs
```diff
@@ -110,10 +110,20 @@ impl CBCharacteristic {
 pub enum CoreBluetoothReply {
     ReadResult(Vec<u8>),
     Connected(BTreeSet<Characteristic>),
+    Disconnected,
+    State(CoreBluetoothState),
     Ok,
     Err(String),
 }
 
+#[derive(Clone, Debug)]
```
qwandor, line 119: Nit: Might as well derive Copy, Eq and PartialEq too.

qwandor (commented):
Thanks for the PR! Can you please rebase it against the dev branch? Most development happens there.

PR #210, opened 2021-09-09
winrtble: Cache is no longer used when requesting gatt services in `connect`

urholaukkarinen:
`BLEDevice::connect` used a cache when requesting gatt services.
This means that a connection to the device might not be established if the services were previously cached.
Thus, a peripheral may not always be connected even if `Peripheral::connect` returns `Ok`.

After this change the gatt services are requested without the cache when connecting. `discover_services` may still use the cache.

### DIFF
--- a/src/winrtble/ble/device.rs
+++ b/src/winrtble/ble/device.rs
@@ -16,7 +16,9 @@ use crate::{api::BDAddr, winrtble::utils, Error, Result};
 use bindings::Windows::Devices::Bluetooth::GenericAttributeProfile::{
     GattCharacteristic, GattCommunicationStatus, GattDeviceService, GattDeviceServicesResult,
 };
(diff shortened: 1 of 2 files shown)


src/winrtble/ble/device.rs
```diff
@@ -58,19 +60,36 @@ impl BLEDevice {
...
+            .GetGattServicesWithCacheModeAsync(cache_mode)
+            .map_err(winrt_error)?;
         let service_result = async_op.await.map_err(winrt_error)?;
         Ok(service_result)
     }
 
     pub async fn connect(&self) -> Result<()> {
-        let service_result = self.get_gatt_services().await?;
+        if self.is_connected().await? {
+            return Ok(());
```
urholaukkarinen, line 78: Peripheral's `connected` value is never set to true if there already was a connection. It probably should be.

- [ ] set `connected` to true when this returns `Ok(())`?

qdot (approved):

PR #211, opened 2021-09-11
Fix reference cycles in peripherals

urholaukkarinen:
AdapterManager and Peripherals had reference counted pointers to each other. This prevented them from ever getting disposed. If a connected peripheral is never disconnected manually, it would stay connected until the process is terminated.

There is now ever only a single instance of AdapterManager and peripherals have a weak reference to it.

Additionally, there was a reference cycle between Peripheral and BLEDevice in winrtble. ConnectedEventHandler in BLEDevice now uses a weak reference.

### DIFF
--- a/src/common/adapter_manager.rs
+++ b/src/common/adapter_manager.rs
@@ -17,20 +17,14 @@ use dashmap::{mapref::one::RefMut, DashMap};
 use futures::stream::{Stream, StreamExt};
 use log::trace;
 use std::pin::Pin;
-use std::sync::Arc;
 use tokio::sync::broadcast;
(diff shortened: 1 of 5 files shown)


qdot:
LGTM, but I'm gonna have @qwandor take a look at it too just because I might've missed something on the ownership changes in AdapterManager.

qdot:
@urholaukkarinen Looks like there's a conflict between this and the UWP cache PR I already pulled in. Can you take a look real quick and make sure the connected status is still updated correctly, if needed?

qwandor (approved):

PR #217, opened 2021-10-20
Clean up PeripheralId on Linux

qwandor:
Use `DeviceId` rather than MAC address as `PeripheralId` on Linux.

This is purely an internal cleanup with no API impact.


qdot:
I mean, an even bigger question: Is anyone using the serialization stuff in the first place?

qwandor:
Fixed the serialisation issue, so this now builds and passes tests.

qdot (approved):

qdot:
Still wondering if anyone actually uses serialization, but lgtm

PR #222, opened 2021-10-31
more efficient From<&MacAddress> for BDAddr

icewind1991:
No need to create and parse a string when we can just copy the bytes around


qwandor:
This is already fixed in the dev branch.

PR #226, opened 2021-11-25
chore: Update to windows-rs v0.28

qdot:
Removes the binding generator in favor of pre-generated APIs, so
we have to define everything in Cargo.toml. Also changes some
type/module locations. Since we're not generating bindings now,
we also can't create From<T> traits on things, so had to change
one From<T> impl to a conversion method.


qwandor (approved):
Looks good, does this mean that we can build the Windows target on other platforms now?

qdot:
I think so? Not real sure, haven't tried it yet, but I think all of the generated code we need is available in the crate now. Are you thinking about cross compiling or something?

PR #228, opened 2021-11-29
Made PeripheralId always be convertable from and to BDAddr

VictorKoenders:
On Windows, there is a conversion from `BDAddr` to `PeripheralId`. However there's no way to convert it back to a `BDAddr`. These conversions don't exist on linux or mac.

I'm making an abstraction layer for a bluetooth device that can run on both windows/linux/mac, as well as embedded. For this I have my own `PeripheralId` that I use internally. `btleplug` will handle the `std` implementation, and [rubble](https://docs.rs/rubble) will be the embedded implementation.

For this abstraction layer I need to convert my own `PeripheralId` type into btleplug's one. The best way I can see this happening is to have a conversion from/to:
- BDAddr
- [u8; 6]
- Something else?

The BDAddr conversion is already in place for Windows, so I figured I'd pick that route. Let me know if there's a different way that might make more sense.

### DIFF
--- a/src/bluez/peripheral.rs
+++ b/src/bluez/peripheral.rs
@@ -251,6 +251,18 @@ impl From<&MacAddress> for PeripheralId {
     }
 }
 
+impl From<BDAddr> for PeripheralId {
+    fn from(address: BDAddr) -> Self {
+        PeripheralId(address)
+    }
+}
+
+impl Into<BDAddr> for PeripheralId {
+    fn into(self) -> BDAddr {
+        self.0
+    }
+}
+
 impl From<bluez_async::AddressType> for AddressType {
     fn from(address_type: bluez_async::AddressType) -> Self {
         match address_type {
--- a/src/corebluetooth/peripheral.rs
+++ b/src/corebluetooth/peripheral.rs
@@ -365,3 +365,16 @@ impl From<SendError> for Error {
         Error::Other("Channel closed".to_string().into())
     }
 }
+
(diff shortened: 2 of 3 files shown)


VictorKoenders:
Looks like apple is not following the spec. As per [their documentation](https://developer.apple.com/documentation/corebluetooth/cbuuid)

> The Bluetooth Special Interest Group (SIG) publishes a list of commonly-used UUIDs, many of which are 16- or 32-bits for convenience. The CBUUID class provides methods that automatically transform these predefined shorter UUIDs into their 128-bit equivalent UUIDs. 

So apple should be using `[u8; 6]` instead of `Uuid`. I'm not sure what their internal implementation is to stretch the value by another 10 bytes, but for now I've just implemented `TryFrom` and `TryInto` which will always fail.

Unfortunately this means that apple has a different implementation from the other platforms, but this is something we can work around for now and maybe fix in the future

qwandor:
CoreBluetooth does not expose MAC addresses to applications, I guess for some kind of privacy reasons. The OS instead assigns random UUIDs to peripherals which applications can use to identify them. These are nothing to do with the service or characteristic UUIDs used by GATT. It doesn't make sense to convert them to MAC addresses, which is why we have the `PeripheralId` type to abstract across peripheral identifiers across different platforms.

I'm not sure why you want to convert your own `PeripheralId` type to btleplug's one. I suggest you just have your type wrap the btleplug `PeripheralId` when building on a platform which uses btleplug, and contain whatever else you want on the other platforms.

Issue #224, opened 2021-11-02
Bluetooth discovery doesn't work on macOS Monterey


gbaranski:
**Describe the bug**
Bluetooth discovery doesn't work on MacOS Monterey

**Expected behavior**
Scanning does discover devices.

**Actual behavior**
Scanning doesn't discover any devices. 

**Additional context**
Running on macOS Monterey. Tested using few terminal emulators, permissions are granted. It works on macOS Big Sur. Probably related to https://github.com/hbldh/bleak/issues/635 and https://stackoverflow.com/questions/69661809/no-name-or-address-cbcentralmanager-no-longer-working-on-macos-12

subtle-supernova:
I've confirmed this is still an issue in MacOS 12.1. I tried to pass a UUID to `ScanFilter` similar to what was outlined here: https://github.com/hbldh/bleak/commit/5e1a9b334f5a72dc7d59aa398230fa997dca1c40 with something like

```rust
let scan_filter = ScanFilter{services: vec!(uuid_from_u16(0xFDCE))};
```

Of course I could have done that way wrong as this is my first time doing anything with bluetooth in my life, but decided to add this here in case someone stumbles upon a working solution using `ScanFilter` or this is useful in some way.

qdot:
afaik we don't actually use the scanfilter for anything but WebBluetooth right now. We'll need to add it to macOS support I think.

psifertex:
FYI, based on https://github.com/hbldh/bleak/issues/720#issuecomment-1009198158

I gave it a try by creating a signing certificate via key chain / create self-signed code signing cert

Then:

codesign -f -o runtime --timestamp -s "Signing Certificate Name" Bundle.app

And then it works fine!

metsis:
This seems to work again in macOS Monterey 12.3.

qdot:
This works fine on Monterey and is being distributed as part of programs now, closing.

PR #229, opened 2021-12-18
initial work on bluetooth from macos 12.x using filter

subtle-supernova:
Initial work on a solution for #224 

Hi,
I'd like to preface this by stating I'm super new to bluetooth, objective-c ffi, rust, and open source. So I'm about as in over my head as it gets. But I had a test environment for this bug and a desire to fix it, so here we are. I figured even if this doesn't end up being a good patch, it's a good test. I've tested this against [this gist](https://gist.github.com/subtle-supernova/495e9f0931b132751f85d2c109486fed) which scan for sennheiser devices and therefor pickup my headphones. My test returned a device id `00:00:00:00:00:00` with the connected bool as `false`, but I'm unsure if that's related to any of the changes I've made here.

Please let me know if there's anything else I can do, and thank you for the library.

### DIFF
--- a/src/corebluetooth/framework.rs
+++ b/src/corebluetooth/framework.rs
@@ -70,9 +70,7 @@ pub mod ns {
     }
 
     pub fn string_to_cbuuid(s: *mut Object /* NSString */) -> *mut Object /* CBUUID */ {
-        unsafe {
-            msg_send![Class::get("CBUUID").unwrap(), UUIDWithString:s]
-        }
+        unsafe { msg_send![Class::get("CBUUID").unwrap(), UUIDWithString: s] }
     }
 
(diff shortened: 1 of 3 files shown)


matoushybl:
I can confirm that this change indeed workarounds the problem with scanning, I am able to find my device with artificially added service solely for this purpose, however the underlying bug is still not fixed on latest Monterey - scanning with empty service UUIDs should work.

qwandor:
I've cleaned up the code a bit. Can you let me know if it still works fine for you? If so I'll get this merged.

qwandor (approved):

matoushybl:
Thanks, works after the refactoring!

PR #233, opened 2021-12-23
In docs use events() instead of event_receiver().

ratmice:
Minor thing that AFAICT `events_receiver()` is gone, in favor of `events()`.
This patch just updates places in docs and examples where `events_receiver()` was referred to.


qwandor:
Thanks for the PR! Can you please rebase it and send it against `dev` rather than `master`? Or I'll do so in a bit.

qwandor (approved):

PR #236, opened 2021-12-26
use select!, stream_map in event_driven_discovery.

ratmice:
This uses StreamMap to make the event_driven_discovery example work with multiple adaptors,
and uses tokio::select! instead of the while loop.

I also added an assertion to show how to get a peripheral from the identifier.
Feel free to close this however, if you think it is getting into territory beyond that fit for examples.

### DIFF
--- a/examples/event_driven_discovery.rs
+++ b/examples/event_driven_discovery.rs
@@ -1,15 +1,12 @@
 // See the "macOS permissions note" in README.md before running this on macOS
 // Big Sur or later.
 
-use btleplug::api::{bleuuid::BleUuid, Central, CentralEvent, Manager as _, ScanFilter};
-use btleplug::platform::{Adapter, Manager};
+use btleplug::api::{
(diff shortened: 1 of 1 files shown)


qwandor:
Thanks for the PR.

I think given that having more than one adapter on one machine is a fairly unusual special case (and only supported by btleplug on Linux in any case), the extra complexity in the example is more confusing than helpful. If you think it is helpful to have an example of using multiple adapters simultaneously is useful, then I think it's better to make a new example rather than modifying this existing one. (The new example could be basically what you have here, with a comment at the top explaining that it's not usually necessary, and only on Linux.)

ratmice:
I think if it is only supported on linux, It might be better to not advertise it with an example, so I think i'll close this for now. But can reconsider if that ever changes, or anyone else chimes in that it was useful for them?

qdot:
To fill in a bit of info: Windows will just not handle multiple adapters, it disabled anything after the first by default. I found this out last week. :|

PR #238, opened 2021-12-31
Clean up CoreBluetooth code and use more wrappers from cocoa crate

qwandor:
I've removed a few unused functions, used some of the wrappers and types from the `cocoa` crate (seeing as we are depending on it anyway) and some of the helpers we already have. This simplifies some of the code a bit.


qdot:
This all looks great, thanks!

qdot (approved):

PR #239, opened 2022-01-04
add a trait PeripheralIdent, and a hashing mechanism for Peripheral.

ratmice:
This adds a newtype around Peripheral, and moves the id() function of
the Peripheral trait to a new trait PeripheralIdent.

PeripheralIdent is implemented for CentralEvent, Peripheral, PeripheralId,
and PeripheralIdent.

A newtype PeripheralIdKeyed which has a Peripheral is added,
This implements Hash, and Borrow for PeripheralId.

So you can test an event/id/peripheral against a PeripheralIdKeyed in
a HashMap.

see the examples/hashing.rs in the commit.

### DIFF
--- a/examples/hashing.rs
+++ b/examples/hashing.rs
@@ -1,0 +1,62 @@
+// See the "macOS permissions note" in README.md before running this on macOS
+// Big Sur or later.
+
+use btleplug::api::{Central, CentralEvent, Manager as _, PeripheralIdent, ScanFilter};
+use btleplug::platform::{Adapter, Manager, Peripheral, PeripheralId, PeripheralIdKeyed};
+use futures::stream::StreamExt;
+use std::collections::HashSet;
+use std::error::Error;
+
+async fn get_central(manager: &Manager) -> Adapter {
+    let adapters = manager.adapters().await.unwrap();
+    adapters.into_iter().nth(0).unwrap()
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    pretty_env_logger::init();
+    let manager = Manager::new().await?;
+    let central = get_central(&manager).await;
+    let mut events = central.events().await?;
+    let mut ids = HashSet::<PeripheralId>::new();
+    let mut keys = HashSet::<PeripheralIdKeyed>::new();
+    central.start_scan(ScanFilter::default()).await?;
+
+    while let Some(event) = events.next().await {
+        match &event {
+            CentralEvent::DeviceDiscovered(id) => {
+                let periph = central.peripheral(&id).await?;
+                assert!(event.id() == periph.id() && id.id() == *id);
+                ids.insert(event.id());
+                // Peripheral doesn't implement Hash, but PeripheralIdentKey does,
+                // It hashes a periheral based upon the hash of its PeripheralId.
+                //
+                // If the platform returns the same PeripheralId for peripherals on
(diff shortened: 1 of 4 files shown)


qwandor:
It's not clear to me what advantage this gives. What are you trying to achieve with these new types and traits?

ratmice:
Just putting a Peripheral into a Hash, and looking up by a `PeripheralId`, 
basically just what is in the `examples/hashing.rs`.

Should also work with `Ord`, if we can add `Ord` for `PeripheralId` in a follow up patch.
I don't think there is anything here which can't be done via a `newtype` outside of btleplug though modulo some additional cloning.

qwandor:
Okay, then two questions:

1. Would it be enough just to implement `Hash` and `Ord` for `PeripheralId`? Then you could have a `HashMap<PeripheralId, Peripheral>` and do the lookup that way. This could be done without the newtype wrapper or extra trait.
2. Why not just use `Central::peripheral` to get the peripheral by ID when you need it?

ratmice:
1. `HashMap` is already implemented for `PeripheralId`, so we can already do `HashMap<PeripheralId, Peripheral>`
But I have another `HashMap<PeripheralId,DeviceState>` cruically the value type of that `HashMap` has trait bounds which `Peripheral` does not have, I can look into seeing if it is possible to turn this into a tuple with `HashMap<PeripheralId, (Peripheral, DeviceState)>` with the trait bounds on the second element.

I'll have to think about whether the latter is possible, because of traits I think I'd have to use MyTuple<Peripheral, DeviceState>.
Implementing those trait bounds implemented for `DeviceState` on MyTuple.  (Since I don't own the trait or rusts tuple).

2. in this case `Central::peripheral` I believe may return a device which has gone away, and we may want to turn on scanning and rescan for it.

As I said in my first commit, this is just an experiment, It might be that there is a clean way I can avoid the Trait bound issues,
This was just a first stab at getting rid of multiple `HashTable<PeripheralId, ...>`, tables which cause multiple lookups etc.

ratmice:
FWIW, I wasn't able to integrate *most* of the API in this patch in my actual program either, as I ended up needing to newtype to implement traits on both key and value for unforeseen reasons.

The only thing from this I did end up using was the get_id() function, which wasn't actually necessary to use,
but get_id(), is the only thing I couldn't actually implement from outside the library.

qwandor:
So you're just saving a clone of the `PeripheralId`? I'm inclined to drop this to be honest, but what we can do is guarantee that `PeripheralId` implements `Hash` and `Ord` on all platforms.

qwandor:
I've sent #240 to do that.

ratmice:
Yep, I'm inclined to drop it too, would you like a patch for getting a ref of a `PeripheralID` from `Peripheral` w/o the extra trait?
That is probably what I should have asked earlier.  

Thanks for the implementation of `Ord`.

qwandor:
If you really want, but honestly `PeripheralId` is not that expensive to clone (especially compared to the cost of talking to Bluetooth hardware) so I'm not sure it's worth optimising.

ratmice:
okay, I'll keep using it locally and once my program settles, I'll look and see what the damage is.
But as i've been writing it, given `move |..|` closures and their relative inflexibility in describing what exactly to move.
I've been just trying to limit owned values to places that *need* ownership, 

It's more that it sometimes enables the code to not have to introduce temporary variables...

PR #248, opened 2022-02-11
build: bump `dashmap` version to `5.1.0` and resolve `RUSTSEC-2022-0002`

antoinevg:
The [RUSTSEC-2022-0002](https://rustsec.org/advisories/RUSTSEC-2022-0002.html) advisory identifies a memory vulnerability in version `5.0.0` of the [`dashmap`](https://crates.io/crates/dashmap) crate.

This vulnerability has been resolved in `dashmap 5.1.0`

More information:

* https://github.com/xacrimon/dashmap/issues/167
* https://github.com/xacrimon/dashmap/pull/180

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ thiserror = "1.0.30"
 uuid = "0.8.2"
 serde_cr = { package = "serde", version = "1.0.133", features = ["derive"], default-features = false, optional = true }
 serde_bytes = { version = "0.11.5", optional = true }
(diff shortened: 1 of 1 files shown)


thomcc:
Hm, those are semver-compatible so we should just be able to `cargo update` to get it resolved on our end, regardless of what they do.

I think perhaps part of the issue was either a stale lockfile (which will keep us on an earlier version), or just that the update to mark the advisory as patched hadn't gone through

qwandor (approved):

PR #258, opened 2022-04-18
Add Display and DeviceID impl.

DD5HT:
I added a Display implementation and the ability to access the device id.
Should I implement similar methods for the other Operating Systems?

Actually I'm not sure if we should expose the device id directly, maybe the path is enough?
What I actually want is the BDAddr here, (like it's done in the Windows API) exposing it here would be great.


qwandor:
If you want the `BDAddr` you can just call the `Peripheral::address` method.

DD5HT:
@qwandor  But I only have the `PeripheralId` available and not the `Peripheral` when using eg `ServiceDataAdvertisement`

PR #264, opened 2022-09-19
Implement disconnect for CoreBluetooth

qdot:
Implementing disconnect for CoreBluetooth. Pretty much just glue code for CB calls that were already in the library.

@qwandor Reviewing mostly as ritual/habit here, can't remember how familiar you are with the CB stuff, so nbd if you aren't up to reviewing it. I've done some cursory testing with this on iOS and it seems happy.

I need this for my upcoming mobile app release so hoping to land/release soon.

### DIFF
--- a/src/corebluetooth/internal.rs
+++ b/src/corebluetooth/internal.rs
@@ -140,6 +140,7 @@ struct CBPeripheral {
     pub peripheral: StrongPtr,
     services: HashMap<Uuid, ServiceInternal>,
     pub event_sender: Sender<CBPeripheralEvent>,
+    pub disconnected_future_state: Option<CoreBluetoothReplyStateShared>,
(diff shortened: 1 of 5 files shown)


src/corebluetooth/internal.rs
```diff
@@ -229,6 +232,15 @@ impl CBPeripheral {
...
         }
     }
+
+    pub fn confirm_disconnect(&mut self) {
+        self.disconnected_future_state
+            .take()
+            .unwrap()
+            .lock()
+            .unwrap()
+            .set_reply(CoreBluetoothReply::Disconnected);
```
qwandor, line 242: Could you just use `CoreBluetoothReply::Ok` here, rather than introducing a new variant?

qwandor (approved):

PR #266, opened 2022-09-21
Make peripheral id field public

Yohannfra:
I need to access this field for my project. I don't see any disadvantage in making it public.

Thanks for making and supporting this crate :)

### DIFF
--- a/src/bluez/peripheral.rs
+++ b/src/bluez/peripheral.rs
@@ -32,7 +32,7 @@ struct ServiceInternal {
     serde(crate = "serde_cr")
 )]
 #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct PeripheralId(pub(crate) DeviceId);
+pub struct PeripheralId(pub DeviceId);
 
 /// Implementation of [api::Peripheral](crate::api::Peripheral).
 #[derive(Clone, Debug)]
--- a/src/corebluetooth/peripheral.rs
+++ b/src/corebluetooth/peripheral.rs
@@ -45,7 +45,7 @@ use uuid::Uuid;
     serde(crate = "serde_cr")
 )]
 #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct PeripheralId(Uuid);
+pub struct PeripheralId(pub Uuid);
 
 /// Implementation of [api::Peripheral](crate::api::Peripheral).
 #[derive(Clone)]
--- a/src/winrtble/peripheral.rs
+++ b/src/winrtble/peripheral.rs
@@ -52,7 +52,7 @@ use windows::Devices::Bluetooth::{Advertisement::*, BluetoothAddressType};
     serde(crate = "serde_cr")
 )]
 #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct PeripheralId(BDAddr);
+pub struct PeripheralId(pub BDAddr);
 
 /// Implementation of [api::Peripheral](crate::api::Peripheral).
 #[derive(Clone)]


qwandor:
The disadvantage of making this public is that it is different on each platform, so client code will need to handle each platform differently. It is supposed to be an implementation detail and not part of the API. What are you trying to do that requires direct access to the details of the `PeripheralId`?

Yohannfra:
Because I need to print this PeripheralId and if I make it public I can just use 

```rs
            let u: PeripheralId = p.id();
            println!("{}", p.id().0.to_string());
```

Without it public I must to a dirty trick to print it
```rust
            let mut uuid_str = format!("{:?}", p.id());
            uuid_str.pop().unwrap();
            let uuid_str = &uuid_str[13..];

```

Because I don't want 
```
PeripheralId(ec7be9c3-2fb7-8cc3-88f4-d6d672a53b59)
```
but just
```
ec7be9c3-2fb7-8cc3-88f4-d6d672a53b59
```

Yohannfra:
I don't understand why is the id different for windows or linux ? It's the peripheral id right ? it should be the same everywhere no ?

Anyway, any solution, **Display**, **to_string** or **pub** would work and be useful I think.

qwandor:
The `PeripheralId` is just a handle that identifies it to the OS, it's not anything fundamental to the peripheral. MacOS assigns arbitrary UUIDs, Windows uses the Bluetooth address (something like a MAC address), and Linux uses the D-Bus path (which contains the adapter ID and Bluetooth address). As far as the btleplug API is concerned, it's just a handle you can use to refer to the peripheral.

If you want something more consistent between platforms you could use the Bluetooth address, which you can get by calling `Peripheral::address`, but MacOS doesn't expose that so we just return a dummy value instead.

Yohannfra:
So the PeripheralId **IS** the most consistent between platforms even if it's not the same.

** Peripheral::address** being a dummy value on macos makes it unusable in a cross platform program (at least for my usage)

Yohannfra:
Maybe a function like 

```rust
get_peripheral_identifier() -> String
```
That returns the mac address on Linux / Windows and the UUID on mac os would makes sense ? 

I need this feature to identify devices that share the same advertising name and be able differentiate them. 

Thanks you,

qwandor:
If you just want to differentiate devices internally you can use the `PeripheralId`; it implements `Eq`, `Hash` and `Ord`. If you want to show something in a UI, then I don't think the Mac OS UUID makes much sense, because it doesn't really mean anything, it's just an internal implementation detail of the OS.

Yohannfra:
I know it's an internal representation but how do I visually to the user the difference between two peripherals if they have the same name and no ID or anything to differentiate them ?

Maybe I could reword this PR to make public only the OSX version of the UUID, so It's available on OSX only and if anyone want to use it, it's available.

Also for your information, this UUID is shown on almost all ble programs on osx:

Lightblue:

Capture d’écran 2022-09-22 à 14 01 39

 Bluesee:
Capture d’écran 2022-09-22 à 14 02 37


It is also available to the user in the C++ library [SimpleBLE](https://github.com/OpenBluetoothToolbox/SimpleBLE), and certainly many others.

Thanks

Yohannfra:
Hi @qwandor, any news on this ?

Thanks you

paroga:
I'm against making the implementation detail public too, but I see no problem in implementing `as_string` on `PeripheralId` with a **CLEAR** documentation that it works different across platforms. (I'm not sure where you want to add `get_peripheral_identifier`, so maybe it's similar to your proposal). Implementing the [`ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html) on `PeripheralId` might be an option too, but I don't see where to put the documentation of the platform-specific behavior then.

qwandor:
Closing in favour of #268, that should let you do what you want.

Yohannfra:
This is perfect, thanks you @qwandor !

PR #268, opened 2022-10-06
Implement Display for PeripheralId

qwandor:
This lets clients render the `PeripheralId` in some platform-specific way for users to distinguish devices, without exposing the internal representation to the API as #266 proposed to do.


qdot (approved):
lgtm, thanks for handling this!

Issue #273, opened 2022-10-22
"Random" panic in CBPeripheral::confirm_disconnect, likely when device loses connection


mmastrac:
**Describe the bug**

When sending a set of status update requests to a BLE device, the code eventually panics deep within the `CoreBluetooth` interface. I can't reproduce it reliably, but running long enough will result in this crash.

**Expected behavior**

No panic, error message propagates up.

**Actual behavior**

Unexpected panic 

**Additional context**

It's difficult to say exactly what causes this, but in this case I'm simply reading/writing to a characteristic.

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /Users/matt/.cargo/registry/src/github.com-1ecc6299db9ec823/btleplug-0.10.1/src/corebluetooth/internal.rs:238:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/option.rs:755:21
   4: btleplug::corebluetooth::internal::CBPeripheral::confirm_disconnect
             at /Users/matt/.cargo/registry/src/github.com-1ecc6299db9ec823/btleplug-0.10.1/src/corebluetooth/internal.rs:236:9
   5: btleplug::corebluetooth::internal::CoreBluetoothInternal::on_peripheral_disconnect::{{closure}}
             at /Users/matt/.cargo/registry/src/github.com-1ecc6299db9ec823/btleplug-0.10.1/src/corebluetooth/internal.rs:506:13
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
   7: btleplug::corebluetooth::internal::CoreBluetoothInternal::wait_for_message::{{closure}}
[...]

Issue #271, opened 2022-10-21
ERROR btleplug::corebluetooth::peripheral > Event receiver died, breaking out of corebluetooth device loop.


mmastrac:
**Describe the bug**

While trying to connect and disconnect from a Delonghi BLE-enabled coffee maker, the CoreBluetooth event loop error is being shown reliably. I am using a copy of `discover_adapters_peripherals.rs` in my local code for testing.

This bug happens each and every time I try to connect to the machine.

Running Monterey 12.6 (21G115), Intel 2019 Mac.

**Expected behavior**

The code should connect and disconnect reliably.

**Actual behavior**

```
Starting scan on CoreBluetooth...
Peripheral "D1538664" is connected: false
Connecting to peripheral "D1538664"...
Now connected (true) to peripheral "D1538664"...
Discover peripheral "D1538664" services...
Service UUID 00035b03-58e6-07dd-021a-08123a000300, primary: true
  Characteristic { uuid: 00035b03-58e6-07dd-021a-08123a000301, service_uuid: 00035b03-58e6-07dd-021a-08123a000300, properties: READ | WRITE | INDICATE }
  Characteristic { uuid: 00035b03-58e6-07dd-021a-08123a0003ff, service_uuid: 00035b03-58e6-07dd-021a-08123a000300, properties: READ | WRITE }
Disconnecting from peripheral "D1538664"...
 ERROR btleplug::corebluetooth::peripheral > Event receiver died, breaking out of corebluetooth device loop.
```

The `ERROR` line appears to be directly caused by a disconnection attempt, as commenting this line out removes it reliably:

```
                        peripheral
                            .disconnect()
                            .await
                            .expect("Error disconnecting from BLE peripheral");
```

**Additional context**

n/a

Rust-Trends:
@qdot I can second these findings. When I used the code at the bottom on my MacOs (12.6) Mac Book Pro 2018 Intel.

- OK: Removing the disconnect from the code and having my BLE device (iPhone 8) close to it works.
- FAIL: It breaks again if I turn OFF BLE on my Device and the Library calls the `pub fn confirm_disconnect(&mut self)` by it self.
- FAIL: When I then implement #274 
```if let Some(future) = self.disconnected_future_state.take() {
            future
                .lock()
                .unwrap()
                .set_reply(CoreBluetoothReply::Ok)
        }
``` 
It still shows the following, confirm_disconnect0 comment is from me to make sure the BTLEplug was rebuild:

```
[...]

Issue #270, opened 2022-10-19
OSx When a device disappears and you try to connect > Event receiver died


Rust-Trends:
Hi,

System: macOS 12.6
btleplug = "0.10.1"

Working with btleplug scanning devices and connecting to write a characteristic works correctly. Only when I turn off the device, the device is still present in the peripherals list kept by the Manager. When it tries to connect I get the following:

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /Users/bob/.cargo/registry/src/github.com-1ecc6299db9ec823/btleplug-0.10.1/src/corebluetooth/internal.rs:238:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/option.rs:775:21
   4: btleplug::corebluetooth::internal::CBPeripheral::confirm_disconnect
             at /Users/bob/.cargo/registry/src/github.com-1ecc6299db9ec823/btleplug-0.10.1/src/corebluetooth/internal.rs:236:9
   5: btleplug::corebluetooth::internal::CoreBluetoothInternal::on_peripheral_disconnect::{{closure}}
             at /Users/bob/.cargo/registry/src/github.com-1ecc6299db9ec823/btleplug-0.10.1/src/corebluetooth/internal.rs:506:13
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/future/mod.rs:91:19
   7: btleplug::corebluetooth::internal::CoreBluetoothInternal::wait_for_message::{{closure}}
             at /Users/bob/.cargo/registry/src/github.com-1ecc6299db9ec823/btleplug-0.10.1/src/corebluetooth/internal.rs:774:71
[...]

PR #274, opened 2022-10-22
CoreBluetooth: skip fulfilling disconnect future if not present

junes-butt:
When a device disconnects, the `on_peripheral_disconnect` handler fires, which tries to fulfill a future for an awaiting call to `disconnect_peripheral`.

However, a device may disconnect for reasons other than a call to `disconnect_peripheral`. In this case there is no future to fulfill. This leads a panic when unwrapping `disconnected_future_state` in `confirm_disconnect`.
This appears to be the cause of issues #273, #270 and possibly #271.

This change skips fulfilling the future if there is none, to avoid this panic. (The disconnect notification will still fire.)

This seems to work well in my testing, but I'm not familiar with the architecture of btleplug so this may not be the right approach.

### DIFF
--- a/src/corebluetooth/internal.rs
+++ b/src/corebluetooth/internal.rs
@@ -233,12 +233,14 @@ impl CBPeripheral {
     }
 
     pub fn confirm_disconnect(&mut self) {
-        self.disconnected_future_state
-            .take()
-            .unwrap()
-            .lock()
-            .unwrap()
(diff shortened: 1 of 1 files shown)


qwandor (approved):

qwandor:
This whole code probably needs some significant rethinking, but in the meantime this change looks reasonable, once the formatting issue is fixed.

PR #279, opened 2022-11-18
Fix buttplug error in the license.

baconhand:
Found this, uh, _autocomplete_ error in the license.


qdot:
Gosh how could that have ever happened I have no idea I totally did not copypasta the license from a project of that name >.>

PR #282, opened 2022-12-12
Fix two minor typos in doc comment

brookman:


qwandor:
I don't think this is a typo; 'iff' is short for 'if and only if'.

brookman:
I wasn't aware of that abbreviation. Sorry for the troubles.

Issue #212, opened 2021-09-21
C++ Implementatnion in Win10


MackoLysy:
Im trying to connect this lib via c++ in windows but i have some issues.
 - First thing is thtat its dosent disconnect properly.
 -  Second error i that my code throw expetion here:
 ```Rust
manager.unwrap().await;
```
As i check lib implementation you dont check if something goes wrong here
```Rust
impl api::Manager for Manager {
    type Adapter = Adapter;

    async fn adapters(&self) -> Result<Vec<Adapter>> {
        let mut result: Vec<Adapter> = vec![];
        let radios = Radio::GetRadiosAsync().unwrap().await.unwrap();

        for radio in &radios {
            let kind = radio.Kind().unwrap();
            if kind == RadioKind::Bluetooth {
                result.push(Adapter::new());
            }
        }
        return Ok(result);
    }
}
````
Rest part works fine :)

qdot:
4 years old at this point, closing (esp because I ship this under dart/flutter on 5 platforms and it seems to work there now).

PR #283, opened 2022-12-12
Set the "connected" state to false when disconnecting on Windows.

brookman:
After disconnecting from a peripheral with `device.disconnect()` on Windows, `device.is_connected()` still returned `true`. The disconnect event always happened. Only the state was not reflected correctly.

Tested successfully on Windows 11.

This bug is also mentioned in issue #212.


qwandor (commented):
Looks good to me, but maybe @qdot can confirm as I'm not familiar with Windows.

qdot (approved):
lgtm

PR #285, opened 2022-12-12
Improve error propagation in manager.adapters() on Windows.

brookman:
Replace some occurrences of `.unwrap()` with `?` to prevent panics when `Radio::GetRadiosAsync()` fails.
Radios where `radio.Kind()` fails are skipped. I hope that behavior is acceptable.

Problem is mentioned in issue #212.

### DIFF
--- a/src/winrtble/manager.rs
+++ b/src/winrtble/manager.rs
@@ -31,15 +31,12 @@ impl api::Manager for Manager {
     type Adapter = Adapter;
 
     async fn adapters(&self) -> Result<Vec<Adapter>> {
-        let mut result: Vec<Adapter> = vec![];
-        let radios = Radio::GetRadiosAsync().unwrap().await.unwrap();
-
-        for radio in &radios {
-            let kind = radio.Kind().unwrap();
-            if kind == RadioKind::Bluetooth {
-                result.push(Adapter::new());
-            }
-        }
-        return Ok(result);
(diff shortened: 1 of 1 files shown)


src/winrtble/manager.rs
```diff
@@ -31,15 +31,12 @@ impl api::Manager for Manager {
...
-            if kind == RadioKind::Bluetooth {
-                result.push(Adapter::new());
-            }
-        }
-        return Ok(result);
+        let radios = Radio::GetRadiosAsync()?.await?;
+        Ok(radios
+            .into_iter()
+            .filter_map(|radio| radio.Kind().ok())
+            .filter(|&kind| kind == RadioKind::Bluetooth)
```
qwandor, line 38: Nit: could you do this in one filter, like `filter(|radio|radio.Kind() == Ok(RadioKind::Bluetooth))`?

qwandor (commented):
This looks like an improvement over the current code so I think we should merge it, but what happens on a machine with more than one Bluetooth adapter? Does it return a vector of two identical elements? (No need to fix it in this PR, but maybe we should at some point.)

src/winrtble/manager.rs
brookman: Updated. Thanks for the hint.

qdot:
afaik, it's not actually possibly to have > 1 bluetooth adapter on windows 10+. Any adapter added after the first will fail to register with the system.

PR #290, opened 2023-01-09
Set `DuplicateData = true` for bluez

### LINKED ISSUE #289: Devices sending multiple advertisements get filtered on bluez/linux
**Describe the bug**

Devices sending multiple types of advertisements (manufacturer data, service data) get de-duplicated by `bluez` by default, so you only see _new_ advertisement types from a device less than 10% of the time (probably when there's a MAC rotation?).

This is particularly an issue when scanning Android and iOS devices, because they send out advertisements pretty much constantly, so it misses "new" things... so scanning them with `bluez` is pretty much unusable.

An easy way to (provably) trigger this is by using [caBLE (hybrid FIDO/WebAuthn authenticators)](https://android-developers.googleblog.com/2022/10/bringing-passkeys-to-android-and-chrome.html); which you can do on Android devices with a recent version of Chrome and Google Play Services (October 2022), or with iOS 16 or later[^ios] -- scanning such a QR code will cause it to start advertising service data payloads with UUID `0xfde2` or `0xfff9`.

**Steps to reproduce:**

1. Run the `event_driven_discovery` example on Linux / bluez

2. Scan this QR code[^url] with a device running iOS 16 or later[^ios] or [Android with a version of Chrome and Google Play Services from October 2022 or later](https://developers.google.com/identity/passkeys/supported-environments):

   FIDO:/196944614260737660800636836935388126324298063435663648500979208347122763132056627778028993444361869915189377361839923290058725964354330451710836675838722109321447142404

   This will start the "make credential" flow on the device, and cause it to start sending new service data advertisements with UUID `0xfde2` or `0xfff9`.  That should time out after about 60 seconds, because the next step of the protocol requires the initiator to take some data in that service data advertisement and use it to connect to a Websocket tunnel that the device is waiting on.

[^url]: FIDO:/196944614260737660800636836935388126324298063435663648500979208347122763132056627778028993444361869915189377361839923290058725964354330451710836675838722109321447142404
[^ios]: iOS 15 scans it, but doesn't work correctly.

**Expected behaviour**

`ServiceDataAdvertisement` events for `0000fde2-0000-1000-8000-00805f9b34fb` or `0000fff9-0000-1000-8000-00805f9b34fb` after the above QR code is scanned.

**Actual behaviour**

No `ServiceDataAdvertisement` events for those UUIDs (about 90% of the time).

**Additional context**

* Tested this with a fresh Debian 11.6.0 install, which has Linux 5.10.0-20-amd64, bluez 5.33-3.1, rust 1.66.0 (installed from rustup)

* Running the same `event_driven_discovery` example on macOS and Windows works correctly

* Setting `filter.duplicate_data = Some(true)` in `impl Central for Adapter::start_scan()` (in `src/bluez/adapter.rs`) _fixes_ the issue (#290).

  This is despite [the documentation saying `DuplicateData` is set to `true` by default](https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/adapter-api.txt?id=6e49216ad47dc8ffd9cbf2066d13fd7b59613c98#n112); even tracing `btleplug`'s DBus calls with `bustle` shows that it's not setting that field by default; and there's different behaviour.

* If the above QR code stops working (it has a timestamp in it, but clients don't appear to check it), you can get a new one by registering a new WebAuthn credential using Chrome or Safari, then selecting "authenticate with another device", which will display a QR code to scan.  [There's a demo site here which will do it](https://webauthn.firstyear.id.au).

micolous:
Fixes #289

This makes bluez discovery work like it does on macOS and Windows.

### DIFF
--- a/src/bluez/adapter.rs
+++ b/src/bluez/adapter.rs
@@ -59,6 +59,7 @@ impl Central for Adapter {
     async fn start_scan(&self, filter: ScanFilter) -> Result<()> {
         let filter = DiscoveryFilter {
             service_uuids: filter.services,
+            duplicate_data: Some(true),
             transport: Some(Transport::Auto),
             ..Default::default()
         };


qwandor (approved):

qdot:
Trusting @qwandor's since I've stayed blissfully ignorant of bluez so far.

qwandor:
From what I remember setting it to `true`, `false` or nothing all have different behaviour, but I don't remember exactly what. I haven't touched this in ages.

qdot:
Guess we'll find out! :D

micolous:
> From what I remember setting it to `true`, `false` or nothing all have different behaviour, but I don't remember exactly what. I haven't touched this in ages.

I vaguely recall seeing this issue before a few years ago for something else, but don't remember what `bluez` _actually_ does, only that its docs were totally wrong. At the time, making my program talk HCI to the kernel was much easier.

> I've stayed blissfully ignorant of bluez so far.

By comparison, `deviceplug`'s macOS and Windows bindings both work really well; I wrote a bunch of code on macOS, which I tested and got working. Then I compiled it on a Windows box, and everything just worked fine without changing anything in my code. 👍 

`bluez` is just the difficult one...

ckcr4lyf:
FWIW, this fixes receiving advertisements via btleplug for me on Linux (`bluez=5.66-1` , `linux=6.2.9`)

On current published crate (`0.10.4`) , if I subscribe to events I only see `DeviceDiscovered` , but I am receiving the Advertising Data (confirmed via Wireshark, and using [bluer](https://github.com/bluez/bluer/) library)

ckcr4lyf:
[bluer](https://github.com/bluez/bluer/) also sets this to true: https://github.com/bluez/bluer/blob/8ffd4aeef3f8ab0d65dca66eb5a03f223351f586/bluer/src/adapter.rs#L720-L732

PR #302, opened 2023-04-11
corebluetooth: conditionally link against AppKit framework

jhonboy121:
* this fixes compilation of dylib for ios targets since AppKit is only available in macos sdk

### DIFF
--- a/src/corebluetooth/framework.rs
+++ b/src/corebluetooth/framework.rs
@@ -101,7 +101,7 @@ pub mod cb {
     pub type dispatch_queue_attr_t = *const dispatch_object_s;
     pub const DISPATCH_QUEUE_SERIAL: dispatch_queue_attr_t = 0 as dispatch_queue_attr_t;
 
-    #[link(name = "AppKit", kind = "framework")]
+    #[cfg_attr(target_os = "macos", link(name = "AppKit", kind = "framework"))]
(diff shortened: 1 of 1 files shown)


qwandor (approved):

jhonboy121:
@qwandor workflow run failed because of rustfmt, merging this will not be an issue

qwandor:
I just realised that this PR was against the wrong branch, it should have been merged into `dev` rather than `master`.

jhonboy121:
Ugh, my bad. Anyways this is a small change and *should* not pose any problems.

qdot:
BTW I rebased dev on master to fix this, required a force push so you'll need to act accordingly on next pull.

jhonboy121:
> I've been getting complaints about this elsewhere so I'll see about doing a point release with this soon anyways.

Yes, linker throws the error only with dylib/cdylib crate-type.

micolous:
> I've been getting complaints about this elsewhere so I'll see about doing a point release with this soon anyways.

I'd be pretty keen to see a release with #290 too, because that issue makes scanning advertisements pretty much unusable on Linux/bluez. I had pinned the commit, but then unfortunately it was broken by the force-push for this.

qdot:
v0.10.5 is now on crates. :3

Issue #246, opened 2022-02-03
MTU Size Negotiation / Retrieval in the public API


Nyrox:
**Feature Description**

It would be very useful to be able to negotiate or at least get the negotiated transfer size for a peripheral.
We are very aware that getting this right in a cross-platform manner is quite a pain and other libraries like Noble for NodeJS have just straight up chosen to not support this, but even a platform-specific API to retrieve it on some platforms would be better than nothing.

We have a custom serial protocol built on top of a characteristic pair and settling for the minimum MTU of 20 makes some operations unreasonably slow.

Would love some input on this, I am also open to writing some of the code (specifically for corebluetooth) if this is deemed a worthwhile feature. Thank you.

robbym:
I gathered some information regarding this:

- Windows - [GattSession.MaxPduSize](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattsession.maxpdusize?view=winrt-22621)
  - Have not used.
- macOS / iOS - [CBPeripheral.maximumWriteValueLength](https://developer.apple.com/documentation/corebluetooth/cbperipheral/1620312-maximumwritevaluelength)
  - I've used peripheral.maximumWriteValueLength(for: .withoutResponse) reliably in the past. .withResponse has some issues. [See here.](https://stackoverflow.com/a/68389331)
- Linux
  - Does Bluez support this?
- Android - [BluetoothGatt.requestMtu](https://developer.android.com/reference/android/bluetooth/BluetoothGatt#requestMtu(int))
  - The way I understand it and have used it in the past, you call requestMtu with the largest you can support (e.g. 512), and it will negotiate the min(host.maxMtu, periph.maxMtu) and call onMtuChanged with the new mtu size.

Seems like Windows and macOS/iOS both automatically negotiate, whilst Android requires you to request it. We could have a bit of code that performs the request on the Android version automatically so that the btleplug interface could be consistent. No clue what to do for Linux.

DCNick3:
Bluez seems to support MTU retrieval since version [5.62](https://github.com/bluez/bluez/releases/tag/5.62): https://github.com/bluez/bluez/commit/1abf1400fa18e70a9085be7864f68f7c054fca14

robbym:
I find it kind of silly that they implemented it for a feature that has no wide support yet, and they neglect implementing it on the device level that every existing API uses and all peripheral devices < Bluetooth 5.2 uses.

[...]

PR #305, opened 2023-05-06
Implemented MTU negotiation

robbym:
Issue: https://github.com/deviceplug/btleplug/issues/246


robbym:
Only Windows so far.

MnlPhlp:
Is there any reason to not just use this internally to request the max mtu on Android?
Then report it like in #305.
I would need to figure out how to to get the actual set mtu on Android for that.

robbym:
Closing because I can't be bothered to rebase. Will start anew.

Issue #276, opened 2022-10-29
RSSI and tx_power_level are None on OSX (corebluetooth)


Yohannfra:
**Describe the bug**
The **rssi** and **tx_power_level** properties is set to None in the code. in *src/corebluetooth/peripheral.rs*
```rust
        let properties = Mutex::from(PeripheralProperties {
            address: BDAddr::default(),
            address_type: None,
            local_name,
            tx_power_level: None,
            rssi: None,
            manufacturer_data: HashMap::new(),
            service_data: HashMap::new(),
            services: Vec::new(),
        });
```

**Expected behavior**
Actual rssi and tx_power_level.

**Actual behavior**
No rssi and tx_power_level

Thanks you

ztroop:
I'm currently developing a tool that depends on this library. `0.11.5`, on a Mac (M2). I notice I only get `Option -> None` with `tx_power_level` in every case. It would seem this is still an issue. However, I do see values for `rssi`

qdot:
> *This was generated by AI during triage.*

Closing this issue because CoreBluetooth now populates RSSI from advertisements and supports RSSI reads, and extracts tx power when the platform provides it. These values remain optional when the peripheral or operating system does not expose them.

PR #306, opened 2023-05-22
update the peripheral rssi value in case of advertisement

xqp:
Hey mates,

for my hobby project i want to use an macbook for distance measuring with rssi.

the corebluetooth implementation doesn't support this at this time because the rssi property would never updated.
see: https://github.com/deviceplug/btleplug/issues/276
        https://github.com/deviceplug/btleplug/issues/76
Cheers


qwandor (approved):
Thanks!

PR #307, opened 2023-05-24
Add RSSI support to CoreBluetooth

mgiachetti:
This adds support for RSSI to CoreBluetooth.
Please let me know if it need changes, happy to do it.
Tested on Mac, not on iOS.


qwandor:
Hi, did you see #306? It's merged in the dev branch now. This looks quite similar. Can you check that and see if there is anything you think needs changing?

mgiachetti:
Thanks @qwandor! It's definitely similar and even better than this one.

PR #308, opened 2023-05-26
add: PeripheralId <-> Uuid conversion for macos

ActuallyHappening:
I need to have a cross-platform way of serialising a PeripheralId (or equivalent structure),
so that I can have some identity to the BLE devices I am connecting to.

I am building a front-end to consume connection states.

### DIFF
--- a/src/corebluetooth/peripheral.rs
+++ b/src/corebluetooth/peripheral.rs
@@ -382,6 +382,12 @@ impl From<Uuid> for PeripheralId {
     }
 }
 
+impl From<PeripheralId> for Uuid {
+		fn from(id: PeripheralId) -> Self {
+				id.0
+		}
+}
+
 impl From<SendError> for Error {
     fn from(_: SendError) -> Self {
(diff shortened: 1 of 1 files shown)


ActuallyHappening:
Actually, upon looking at the other platform's it seems that this interoperability of `PeripheralId` and `Uuid::uuid` does not hold, at least it doesn't work for the Linux implementation (Bluez)

Is there some way of safely converting the general (i.e. cross platform) `btleplug::platform::PeripheralId` into a `Uuid::uuid`? I am only on macOS so can't test any other platform's implementation

qdot:
This isn't possible. Only macOS and WebBluetooth convert bluetooth ids to uuid, every other platform uses the bluetooth address as a bluetooth ID. Not only that, macOS/WebBluetooth UUIDs are unique per machine/domain respectively.

ActuallyHappening:
Yes I agree, I am moving to using BDAddr (from btleplug::api) because it is available on wasm32 targets (even though webbluetooth is not yet implemented)

PR #317, opened 2023-06-16
added request_mtu function for android Peripheral

MnlPhlp:
I implemented a request_mtu function that simply requests the maximum mtu on Android. In my case I used it as follows
```rs
#[cfg(target_os = "android")]
{
    peripheral.request_mtu().await?;
}
```
Actually, the function should be implemented as a future waiting for the [onMtuChanged callback](https://developer.android.com/reference/android/bluetooth/BluetoothGattCallback#onMtuChanged(android.bluetooth.BluetoothGatt,%20int,%20int)), but I couldn't get that to work.
The callback was never called on the android side. My data rate was much higher after the request_mtu call, so the mtu was still set.

I wanted to open this PR anyway because this change made the library usable for my project. Without this change, it can't be used on Android if you need a certain data throughput.

Maybe someone with more knowledge about android can improve this, but it is a working basis.

### DIFF
--- a/src/droidplug/java/src/main/java/com/nonpolynomial/btleplug/android/impl/Peripheral.java
+++ b/src/droidplug/java/src/main/java/com/nonpolynomial/btleplug/android/impl/Peripheral.java
@@ -85,11 +85,11 @@ class Peripheral {
     }
 
     
-    public void requestMtu() {
+    public void requestMtu(int mtu) {
         synchronized (this) {
             if (this.gatt != null) {
                 try {
-                    this.gatt.requestMtu(517);
+                    this.gatt.requestMtu(mtu);
                 } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


trik:
The requested mtu value is currently hardcoded, it should be better to take a parameter

MnlPhlp:
Right makes sense. I just changed that

qdot:
Ok, this conflicts with the API in #305, so we're gonna have to figure out some sort of middle ground here. Also need to figure out what it'll look like on CoreBluetooth and Bluez (if it's even possible there).

MnlPhlp:
Is there any reason to not just use this internally to request the max mtu on Android?
Then report it like in #305.
I would need to figure out how to to get the actual set mtu on Android for that.

Issue #325, opened 2023-07-26
往Gatt characteristic写入数据时会将通道阻塞住


Henxun:
**Describe the bug**

**Expected behavior**

**Actual behavior**
A clear and concise description of what actually happened.

**Additional context**

Henxun:
let operation = self
            .characteristic
            .WriteValueWithOptionAsync(&writer.DetachBuffer()?, write_type.into())?;
        let result = operation.await?;

it's block on "let result = operation.await"

Henxun:
当get_characteristics中调用GetCharacteristicsWithCacheModeAsync中传入参数BluetoothCacheMode::Uncached时会阻塞代码运行

qdot:
This should be addressed by commit 758c10e — both `GetCharacteristicsWithCacheModeAsync` and `GetDescriptorsWithCacheModeAsync` now use a 5-second timeout on uncached mode, falling back to cached mode with a warning log if the uncached call hangs. This avoids the indefinite hang while still preferring fresh GATT data when possible.

bpbradley:
> This should be addressed by commit [758c10e](https://github.com/deviceplug/btleplug/commit/758c10e61847d40dca04c7e0bb3f28fa6acba425) — both `GetCharacteristicsWithCacheModeAsync` and `GetDescriptorsWithCacheModeAsync` now use a 5-second timeout on uncached mode, falling back to cached mode with a warning log if the uncached call hangs. This avoids the indefinite hang while still preferring fresh GATT data when possible.

I know this is old, but just resurrecting here where I *think* this is most relevant -- 

But I am noticing that the fix linked here is causing issues on devices which have a modest number of services / characteristics (as in, more than just a few), on Windows. 

Basically, it appears to hang forever. I was able to get it to eventually connect and read services, but it takes multiple minutes for about 15 characteristics.

Briefly looking, it seems the uncached enumeration was timing out, and then falling back to cache mode which would also hang. It seems that the abandoned WinRT operation might hold internal locks that block subsequent GATT operations. So each async read is queueing behind a stuck operation, and so individual characteristic reads were taking ~5 seconds each, adding up to 30+ seconds just to read a handful of values. This adds up pretty quickly if you are trying to read data from a few characteristics.

[...]

PR #327, opened 2023-07-27
Fix code running blocking on windows

Henxun:
#325 https://github.com/deviceplug/btleplug/issues/325


qdot:
I'm iffy on this because every time we've changed cache mode handling on windows in the past, it's broken a *lot* of stuff. I'll have to try it and see what the outcome is.

qdot:
Closing this out as we'll have a method that can handle this without possibly causing breakage.

PR #329, opened 2023-07-28
Fix disconnect when using the winrt backend.

### LINKED ISSUE #300: Windows: disconnect is not working
**Describe the bug**
on Windows, when calling the peripheral.disconnect() to end the connection with a BLE device, it executes successfully, 
but the connection still existed until I exit the main process.
I can receive the DeviceDisconnected event, and everything looks OK, except the BLE device is still connected to my PC.

TristonJ:
Fixes #300 

You must explicitly disconnect from each service before disconnecting from the device. Otherwise, Windows does not seem to fully disconnect the device internally.

See https://stackoverflow.com/questions/39599252/windows-ble-uwp-disconnect for similar issues.

I did this by simply adding a list of services to the `BLEDevice` when `discover_services` is called, allowing us to `Close` them when the device is dropped. This appears to work in my use-case, but I'd be interested to see if it's a robust enough solution for other folks.

### DIFF
--- a/src/winrtble/ble/device.rs
+++ b/src/winrtble/ble/device.rs
@@ -29,6 +29,7 @@ pub type ConnectedEventHandler = Box<dyn Fn(bool) + Send>;
 pub struct BLEDevice {
     device: BluetoothLEDevice,
     connection_token: EventRegistrationToken,
+    services: Vec<GattDeviceService>,
 }
 
 impl BLEDevice {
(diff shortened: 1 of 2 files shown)


thewh1teagle:
Thank you very much!
I can confirm this worked for me in `windows 11`
in `cargo.toml`:
```
btleplug = { git = "https://github.com/TristonJ/btleplug.git", rev = "d90009d0509eb58677b4f106b48b650dd4e13347" }
```

qdot:
Yeah, I think this is fair. Weird that we have to do it, but better than basically leaking live connections.

qdot (approved):

Issue #335, opened 2023-09-06
macOs <12 : `Manager::new()` : CBManager authorization objc error


oletf:
**Describe the bug**

Problem occurring on call of `Manager::new()`

Developing an app on a `12.5.1` macOs where the problem doesn't occur.

I've been able to test on a `10.13.6` where I get a panic (`foreign exception` from rust's POV) 
which I've been able to locate to an [objc crash happening here](https://github.com/deviceplug/btleplug/blob/fd3f092de1f06874272dee5dde222628ea459332/src/corebluetooth/framework.rs#L177) : 
the error being `-[CBManager authorization]: unrecognized selector sent to instance 0x7f9be4601020`

According to [Apple objc SDK documentation](https://developer.apple.com/documentation/corebluetooth/cbmanager/3180033-authorization?changes=latest_major&language=objc) the property went from an `instance property` to a `class property` idrk when exactly, 
but 
- [documentation as linked above](https://developer.apple.com/documentation/corebluetooth/cbmanager/3180033-authorization?changes=latest_major&language=objc)
- [and the header source from the osx11 SDK](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX11.3.sdk/System/Library/Frameworks/CoreBluetooth.framework/Versions/A/Headers/CBManager.h)

both seem to mention that the class property is available since `macOS 10.15+` which I don't have access to for now to verify...

---

Since apparently the  `CBManager` class is available since `10.13` I tried to change the code [from](https://github.com/deviceplug/btleplug/blob/fd3f092de1f06874272dee5dde222628ea459332/src/corebluetooth/framework.rs#L175-L178) : 

```rust
// CBManager
pub fn manager_authorization() -> CBManagerAuthorization {
    unsafe { msg_send![class!(CBManager), authorization] }
}
```
to :
```rust
// CBManager
pub fn manager_authorization() -> CBManagerAuthorization {
    unsafe {
        let cbmanager: id = msg_send![class!(CBManager), new];
        msg_send![cbmanager, authorization]
    }
}
```
to get an instance of the class and try to get the legacy `authorization` property from it.

This change didn't break anything on `12.5`, probably because it still gets the class property through the instance in the end.
But on `10.13` it didn't work better...the call to `new` worked, but the `authorization` didn't.

I discovered objc rust inter-workings with it on the fly, so I probably missed a thing...

[...]

PR #337, opened 2023-09-07
fix CBManager crash on macOS `<10.15`

oletf:
my current fix for #335

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ jni-utils = "0.1.1"
 cocoa = "0.25.0"
 objc = "0.2.7"
 libc = "0.2.147"
+os_info = "3.7.0"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 windows = { version = "0.51.1", features = ["Devices_Bluetooth", "Devices_Bluetooth_GenericAttributeProfile", "Devices_Bluetooth_Advertisement", "Devices_Radios", "Foundation_Collections", "Foundation", "Storage_Streams"] }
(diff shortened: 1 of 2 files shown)


oletf:
actually wondering how format got a fail on a file which is not part of the commit :D

qdot:
I'm *really* bad about updating formatting, so it tends to file through. I need to update our CI to warn, not fail, on formatting.

Don't worry about it, I'll try to get the patch in and formatting updated soon.

qdot:
Ok, looking at this again...

The one problem I see here is that we aren't actually checking the OS platform we're looking at the version of, so this check will run on both macOS *and* iOS (I build this library into an iOS application). Do we need to do extra target checks here for knowing if we're running on macOS and iOS? Mainly want to make sure before I merge this, as debugging this on iOS is a pain if anything goes wrong.

oletf:
yeah I thought about this but forgot to post a thing about it !
should I change the match to cover both ios edge-cases too then ?

oletf:
@qdot  I don't have any iOS device to try it on, would it be possible for you to get what a `dbg!(os_info::get())` does print on your iOS device ?
 I'm not sure of what I'd need to match for to check for iOS vs macOS from os_info crate doc...
 Thanks forward

qdot:
At this point, there's only so much backward compat we can handle, and I already let this sit for 2.5 years. I'll update our OS X compat to 10.15 minimum. Closing.

oletf:
had almost forgotten about this one,
seems like a very reasonable decision to me,
thanks for the feedback ! 👍

PR #352, opened 2023-11-30
Add service_uuid to the ValueNotification struct

night-crawler:
Introduced `service_uuid` to `ValueNotification` struct to enhance clarity. This addition allows users to identify the service associated with each notification.

Not sure where I can take from the `service_uuid` in `droidplug`.


qdot:
Implemented in #445

PR #355, opened 2023-12-10
Drop async-trait

matze:
This is a preview and in anticipation of Rust 1.75. But it works well with current nightly either. Note that the public trait API requires the desugared form in order to specify the `Send` bound on the returned future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ path = "src/lib.rs"
 serde = ["uuid/serde", "serde_cr", "serde_bytes"]
 
 [dependencies]
-async-trait = "0.1.74"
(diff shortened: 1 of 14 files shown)


qdot:
There's gonna be an interesting question about how far back w/ rust versions we want the library to be compatible, so I'm hesitant to merge this right after the 1.75 release (currently scheduled for 2023/12/28). That said, I'll definitely let this hang out and am looking forward to the feature going stable.

matze:
One option would be to make the async_trait macro usage be a default feature (for the foreseeable future) and when opted out switch to AFIT. But I can imagine that this looks rather ugly with lots of `#[cfg(…)]`s

Issue #267, opened 2022-09-21
Peripheral names are missing on Windows 11


Cyclip:
**Describe the bug**
When getting a list of peripherals after scanning, all names are `None` even though they should have names. Other properties are available, only `local_name` is missing.

**Expected behavior**
All peripherals with names should have their names discovered in their properties.

**Actual behavior**
All peripherals have their `local_name` property as `None`.

**Additional context**
Using Windows 11 might be a problem. The code I'm using is:
```Rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    let manager = Manager::new().await?;

    // get adapter
    let adapters = manager.adapters().await?;
    let central = adapters.into_iter().nth(0).unwrap();

    // start scanning for devices
    central.start_scan(ScanFilter::default()).await?;
    time::sleep(Duration::from_secs(2)).await;

    // get a list of devices that have been discovered
    let devices = central.peripherals().await?;
    for device in devices {
        // print the address and name of each device
        let properties= device.properties()
            .await
            .unwrap()
            .unwrap();
        
        let names: Vec<String> = properties
            .local_name
            .iter()
            .map(|name| name.to_string())
            .collect();
        
        println!("{}: {:?}", device.address(), names);
    }

    Ok(())
}
```

azimuthdeveloper:
I'm seeing this too. In another library, where the device name seems to function, it looks like it checks in two places to see if it can find the `local_name`. It's in C++ so it's a different language, but possibly helpful to know the `local_name` could be in one of two locations?

https://github.com/woodemi/quick_blue/blob/master/quick_blue_windows/windows/quick_blue_windows_plugin.cpp#L342

```
  auto device = co_await BluetoothLEDevice::FromBluetoothAddressAsync(args.BluetoothAddress());
  auto name = device ? device.Name() : args.Advertisement().LocalName();
```

jcaude:
I"ve got the same issue on Windows 10 with a similar code running as expected on macOS. I will give a look if I can find something to propose (I don't want to connect to the device at this point

jcaude:
This is what I understand about this issue...

1. For some/many devices the local name is broadcast through a SCAN_RESP packet, thus only available during active scans
2. As soon as you add a scan filter, you filtered out the SCAN_RESP thus the device local name. Oddly, in your sample code you don't have any scan filters other than the default.
3. I add a `drop()` on the RW lock, I don't know if it is related .. but now it works and I got device names

[...]

PR #358, opened 2023-12-19
Fix manufacturer data lost issue (see #267)

jcaude:
This is the patch for issue #267, at least for the manufacturer data lost. I have added some drop() after locks, I not sure about this one, I'm used to release locks as soon as they became unnecessary.

### DIFF
--- a/src/winrtble/peripheral.rs
+++ b/src/winrtble/peripheral.rs
@@ -258,7 +258,7 @@ impl Peripheral {
                     id: self.shared.address.into(),
                     services: services_guard.iter().copied().collect(),
                 });
-                drop(service_data_guard);
+                drop(services_guard);
             }
         }
 
--- a/src/winrtble/peripheral.rs
+++ b/src/winrtble/peripheral.rs
@@ -143,26 +143,30 @@ impl Peripheral {
 
                 let mut local_name_guard = self.shared.local_name.write().unwrap();
                 *local_name_guard = Some(name.to_string());
+                drop(local_name_guard);
             }
         }
         if let Ok(manufacturer_data) = advertisement.ManufacturerData() {
             let mut manufacturer_data_guard = self.shared.latest_manufacturer_data.write().unwrap();
 
-            *manufacturer_data_guard = manufacturer_data
-                .into_iter()
-                .map(|d| {
-                    let manufacturer_id = d.CompanyId().unwrap();
-                    let data = utils::to_vec(&d.Data().unwrap());
+            if manufacturer_data.Size().unwrap() > 0 {
+                *manufacturer_data_guard = manufacturer_data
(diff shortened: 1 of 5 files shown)


src/winrtble/peripheral.rs
```diff
@@ -143,26 +143,30 @@ impl Peripheral {
 
                 let mut local_name_guard = self.shared.local_name.write().unwrap();
                 *local_name_guard = Some(name.to_string());
+                drop(local_name_guard);
```
qwandor, line 146: This has no effect, the guard goes out of scope on the next line so would be dropped here anyway.

src/winrtble/peripheral.rs
```diff
@@ -143,26 +143,30 @@ impl Peripheral {
...
+                        (manufacturer_id, data)
+                    })
+                    .collect();
+
+                // Emit event of newly received advertisement
+                self.emit_event(CentralEvent::ManufacturerDataAdvertisement {
+                    id: self.shared.address.into(),
+                    manufacturer_data: manufacturer_data_guard.clone(),
+                });
+                drop(manufacturer_data_guard);
```
qwandor, line 168: Ditto, the guard is about to go out of scope so this has no effect.

src/winrtble/peripheral.rs
```diff
@@ -217,6 +221,7 @@ impl Peripheral {
                     id: self.shared.address.into(),
                     service_data: service_data_guard.clone(),
                 });
+                drop(service_data_guard);
```
qwandor, line 224: Ditto.

src/winrtble/peripheral.rs
```diff
@@ -253,6 +258,7 @@ impl Peripheral {
                     id: self.shared.address.into(),
                     services: services_guard.iter().copied().collect(),
                 });
+                drop(services_guard);
```
qwandor, line 261: Ditto.

src/winrtble/peripheral.rs
```diff
@@ -263,6 +269,7 @@ impl Peripheral {
                 BluetoothAddressType::Random => Some(AddressType::Random),
                 _ => None,
             };
+            drop(address_type_guard);
```
qwandor, line 272: Again, this is just about to go out of scope and be dropped anyway.

jcaude: Ok, I was not sure about that.. I discard all of them.

jcaude: I just commit the update

qdot:
Commit lined cleaned up and merged as [53f5d41](https://github.com/deviceplug/btleplug/commit/53f5d4134fafdf56aeeb53800c47709cdb012a8c)

PR #360, opened 2023-12-28
fix: winrt: Return error when discovering services fails

danielstuart14:
Prior to this commit, the `discover_services` method would return `Ok(())` even if the underlying calls had failed.

This was causing a deadlock on my system when I tried to reconnect to a device that had just been disconnected.


danielstuart14:
Here is a screenshot of where it would get stuck:
image

qdot (approved):

PR #365, opened 2024-02-06
fix: Duplicate Characteristic UUID overwrite each other

blackspherefollower:
If a device (like the LOOB) declares multiple characteristics on a service with the same UUID, the last declared characteristic is used not the first (which is what the LOOB requires us to control).

Ideally we'd have access to them all, but since we can't tell them apart at the higher levels first wins seems better than last wins.

This has been tested on Windows, macOS, Android and Linux

### DIFF
--- a/src/droidplug/peripheral.rs
+++ b/src/droidplug/peripheral.rs
@@ -8,7 +8,6 @@ use crate::{
 use async_trait::async_trait;
 use futures::stream::Stream;
 use jni::{
-    descriptors,
     objects::{GlobalRef, JList, JObject},
     JNIEnv,
 };
@@ -270,7 +269,7 @@ impl api::Peripheral for Peripheral {
 
             for service in list.iter()? {
                 let service = JBluetoothGattService::from_env(env, service)?;
-                let mut characteristics = BTreeSet::new();
+                let mut characteristics = BTreeSet::<Characteristic>::new();
                 for characteristic in service.get_characteristics()? {
                     let mut descriptors = BTreeSet::new();
                     for descriptor in characteristic.get_descriptors()? {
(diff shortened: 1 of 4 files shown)


src/droidplug/peripheral.rs
```diff
@@ -280,18 +280,18 @@ impl api::Peripheral for Peripheral {
...
-                    peripheral_characteristics.push(Characteristic {
-                        service_uuid: service.get_uuid()?,
-                        uuid: characteristic.get_uuid()?,
-                        properties: characteristic.get_properties()?,
-                        descriptors: descriptors,
-                    });
+                    };
+                    // Only consider the first characteristic of each UUID
+                    // This "should" be unique, but of course it's not enforced
+                    if !characteristics.contains(&char) {
```
qwandor, line 291: What if there is another characteristic which has the same UUID but is different in some other way?

blackspherefollower: It's dropped. Ideally, this should never happen, but the reality turns out that there are devices out there that do this.

Right now, the first characteristic is overridden, so in that respect its no worse (this change just picks the first one rather than the last).

I did consider changing the characteristic structures to Map<Uuid, Vec<CharacteristicObj>> (please excuse the pseudocode), but given that all the other methods use the Uuid as the key for performing characteristic operations, that'd mean that they'd all have to be able to take a UUID and an index (99.9% of the time, there's only going to be 1 item in the array). The trade-off between being able to address all characteristics (even with duplicate keys) vs the overhead of the interface changes to support 1 device felt way off to me.

qwandor: How? This code seems to be comparing the characteristic as a whole, not just the UUID, unlike the other platform implementations.

blackspherefollower: That's a good question... Sorry, I was answering for the change as a whole, not the Android diff specifically.

Hmm... the comparator does look like it applies to all members of the struct, so the contains check probably never returns true. I guess the properties and then descriptors would define the order that btleplug presents the characteristics, and I'm guessing that at a higher level they'll get uniqued by UUID. I see the UUID is the only identifier passed across the JNI boundary, and the characteristics are held in an ArrayList on that side, so I think making this set UUID unique (first characteristic wins) will mean that the presented properties match up the the characteristic being addressed.

blackspherefollower: @qwandor I've re-written and re-tested the Android portion of the change

blackspherefollower:
Rebased and building, but still testing: please don't merge (I swear there was a button for making PR's draft/WIP...)

qdot:
I converted it to draft (It's in the "reviewers" section for some reason)

blackspherefollower:
> I converted it to draft (It's in the "reviewers" section for some reason)

Apparently I don't have the required memberships to see the option...

I've tested on Windows/Linux/Android and we're all good.
My attempts to build for macOS have not been so successful - I'm bouncing around either ruby(cocoapods) issues or cert issues in CI

blackspherefollower:
@qdot Apparently the "not a draft any more" button was right at the bottom of the page...

qdot:
macOS seems fine, landing.

qdot (approved):

Issue #368, opened 2024-03-02
Missing Values For Device Address and TX Power


ztroop:
**Describe the bug**
I'm writing a program that uses this library to scan devices. I'm able to scan and detect devices, but I'm getting `00:00:00:00:00:00` for `address` and `None` for `tx_power_level` consistently across 15+ devices. 

```rs
...
    let manager = Manager::new().await?;
    let adapters = manager.adapters().await?;
    let central = adapters.into_iter().next().ok_or("No adapters found")?;

    central.start_scan(ScanFilter::default()).await?;
    let mut events = central.events().await?;

    let mut devices_info = Vec::new();

    while let Some(event) = events.next().await {
        if let CentralEvent::DeviceDiscovered(id) = event {
            if let Ok(device) = central.peripheral(&id).await {
                let properties = device
                    .properties()
                    .await?
                    .unwrap_or(PeripheralProperties::default());
                let name = properties
                    .local_name
                    .unwrap_or_else(|| "Unknown".to_string());
                let tx_power = properties
                    .tx_power_level
                    .map_or_else(|| "N/A".to_string(), |tx| tx.to_string());
                let address = properties.address.to_string();
                let rssi = properties
                    .rssi
                    .map_or_else(|| "N/A".to_string(), |rssi| rssi.to_string());
...
```

**Expected behavior**
Expecting data to be provided by `properties.address` and `properties.tx_power_level`.

**Actual behavior**
Consistent lack of data.

**Additional context**
Using version `0.11.5`.

qwandor:
Is this on Mac OS? Some platforms don't provide this information, in which case there's not much we can do.

ztroop:
Yes, this is a Mac Air (M2). Is there a matrix of what platforms are supported and offer what type of functionality?

ztroop:
Perhaps this is the case for `tx_power_level` but `address` should show something other than 00:00:00:00:00:00, as other BTLE libraries are able to retrieve this information. I suspect there might be going on there.

qdot:
It can't show anything, because CoreBluetooth does not allow access to device addresses.

ztroop:
So I've been experimenting with different versions originally thinking it was a bug. When I downgrade to `0.8`, I see the addresses without issue:

<img width="889" alt="Screenshot 2024-03-03 at 11 08 39 PM" src="https://github.com/deviceplug/btleplug/assets/5018339/5dcb75bb-33af-4b58-9e0a-cd14046be73f">

[...]

PR #371, opened 2024-03-08
Make Shared.uuid Public

ztroop:
As discussed in [!368](https://github.com/deviceplug/btleplug/issues/368):
> Given the lack of a consistent method to retrieve the actual MAC address, is it still possible to access the uuid as in previous versions?

I believe this is the `uuid` of interest - but it's private. Unless there's a compelling reason not to have it public, this is a PR to change that.

### DIFF
--- a/src/corebluetooth/peripheral.rs
+++ b/src/corebluetooth/peripheral.rs
@@ -56,13 +56,13 @@ impl Display for PeripheralId {
 /// Implementation of [api::Peripheral](crate::api::Peripheral).
 #[derive(Clone)]
 pub struct Peripheral {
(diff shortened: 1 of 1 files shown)


qdot:
The compelling reason not to do this is that, when anyone tries to compile their code that uses this patch for anything other than corebluetooth, it will break. This is a cross platform library and requires alignment across platforms for definitions.

PeripheralId is implemented with Display traits on all platforms, and will encase the platform specific identifier for the device, be it BTAddr or UUID. You can use that across platforms for identification.

ztroop:
Understood, thank you.

PR #381, opened 2024-05-01
chore(corebluetooth): Use `objc2` and its framework crates

madsmtm:
[`objc2`](https://github.com/madsmtm/objc2) is the successor to `objc`, which improves the way memory-management is done using the `objc2::rc::Id` type, which is effectively a typed `StrongPtr`. Additionally it provides the `msg_send_id!` macro, which ensures that the Objective-C memory management rules are followed correctly. Finally, it has encoding verification when debug assertions are enabled, which verify the calling convention at runtime, and can catch errors with using an incorrect method signature. The second commit fixes the errors that was found with this.

Compared to `cocoa`, `objc2-foundation` provides type-safe wrappers around each class, which means that you can't accidentally call e.g. `allKeys` on something that is not a dictionary, and it allows us to naturally use `Option`s for nullable values. The third commit replaces the manual interface to Foundation types in `src/corebluetooth/framework.rs` with the types from `objc2-foundation`.

This PR is only the first part of the transition; in the future we will have `objc2-core-bluetooth`, see https://github.com/madsmtm/objc2/pull/608, which should eliminate the need all of `src/corebluetooth/framework.rs`. But in the meantime, this still improves the safety significantly.

Feel free to ask if something is unclear!


qdot:
Wow, thanks so much for this! The macOS portion of the codebase is probably the least understood part (I really just ripped out what Servo had and poked it until it worked), so I'm happy to see some maintenance happening to hopefully make it safer and more maintainable. I'll try to review this and get it in soon.

qwandor (approved):

Issue #206, opened 2021-08-31
Ability to update or reset the discovered Peripherals


GilShoshan94:
# **Goal**
Be able to update the list of discovered peripherals and drop the ones that does not advertised anymore after a certain timeout.
Like in a android phone when you search for Bluetooth headphone, the list get updated and if you turn off your headphone, it will be drop of the list fairly quickly.

# **Context**
I am currently using the polling method due to the "Linux Device Discovery Caveat" in order to discover the available peripherals.
So I am getting my list of adapter and then call `start_scan()`, wait a bit and then call `peripherals()`, very much like it is shown in the example [subscribe_notify_characteristic.rs](https://github.com/deviceplug/btleplug/blob/master/examples/subscribe_notify_characteristic.rs).

But, I am running the "wait a bit and then call `peripherals()`" in a loop in a spawn tokio task.
And the idea was to send to my main task an updated list of available peripherals.

My issue is that  `peripherals()` as its documentation explain (_Returns the list of Peripherals that have been discovered so far. Note that this list may contain peripherals that are no longer available._) never remove peripheral that are not available anymore (For example, when I turn those devices off).

When digging around into the lib it seems to me that the Adapter struct hold the discovered peripherals and never release them.
It seems that my only work around is to completely drop the adapter and recreate a new one from the manager with `.adapters()`. 

# **Feature Description**
A method on the adapter to explicitly clear the internal list of discovered peripheral.

Or even better but might be harder to implement: to add a u8 field to the peripheral struct, I would called it `ttl` for Time To Live, and whenever a peripheral get discover or rediscover, this field would be set to TTL_MAX, for example 5, and whenever the adapter is polled with `peripherals()` call, each peripheral the adapter holds see their ttl decreased by 1, and then they pass a check, if ttl reach 0, the peripheral get removed from the adapter. This way, calls to `peripherals()` will returns the list of Peripherals that have been discovered so far and that their advertising was receive at least in the passed ttl polls. So the list may still contain peripherals that are no longer available, but far less and more just.

qdot:
Closing this issue because the requested explicit reset operation is implemented as `clear_peripherals()` across the supported backends. TTL or automatic expiration would be a separate design question.

PR #382, opened 2024-05-08
Add clear_peripherals method to adapter

danielstuart14:
Add clear_peripherals method to adapter, so that the peripheral list can be cleared after being used.
Without this, the list would keep growing until the whole adapter object is dropped.

The method returns a result so that later if a platform/functionality needs to be fallible the API won't need to be changed.

AFAIK, Linux doesn't support this as the peripheral list comes from bluez directly.

Solves https://github.com/deviceplug/btleplug/issues/206

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -355,6 +355,9 @@ pub trait Central: Send + Sync + Clone {
     /// Add a [`Peripheral`] from a MAC address without a scan result. Not supported on all Bluetooth systems.
     async fn add_peripheral(&self, address: &PeripheralId) -> Result<Self::Peripheral>;
 
+    /// Clear the list of [`Peripheral`]s that have been discovered so far.
+    async fn clear_peripherals(&self) -> Result<()>;
+
     /// Get information about the Bluetooth adapter being used, such as the model or type.
     ///
     /// The details of this are platform-specific andyou should not attempt to parse it, but it may
--- a/src/bluez/adapter.rs
+++ b/src/bluez/adapter.rs
@@ -101,6 +101,12 @@ impl Central for Adapter {
         ))
     }
 
+    async fn clear_peripherals(&self) -> Result<()> {
+        Err(Error::NotSupported(
+            "Can't clear peripheral list on this platform".to_string(),
+        ))
+    }
+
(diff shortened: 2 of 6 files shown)


src/api/mod.rs
```diff
@@ -355,6 +355,9 @@ pub trait Central: Send + Sync + Clone {
     /// Add a [`Peripheral`] from a MAC address without a scan result. Not supported on all Bluetooth systems.
     async fn add_peripheral(&self, address: &PeripheralId) -> Result<Self::Peripheral>;
 
+    /// Clear the list of [`Peripheral`]s that have been discovered so far.
+    async fn clear_peripherals(&self) -> Result<()>;
```
qwandor, line 359: As far as I can see there's no need for this to be async.

src/bluez/adapter.rs
```diff
@@ -101,6 +101,12 @@ impl Central for Adapter {
         ))
     }
 
+    async fn clear_peripherals(&self) -> Result<()> {
+        Err(Error::NotSupported(
```
qwandor, line 105: I'm worried that this will result in people writing programs that crash on Linux as they only tested it on other platforms. What do you think about this just being a silent no-op on Linux? Not being able to clear the peripheral list shouldn't be a big issue in most cases I can think of.

danielstuart14: That sounds good, I'll change it. I think it may be possible to clear the bluez list, but the bluez-async lib doesn't seem to currently support it.

src/api/mod.rs
danielstuart14: I think it is best to keep it async, so that if a new platform (or even a linux solution later) needs it to be async we don't need to change the API. What do you think?

qdot (changes requested):
Making this as changes requested since there was some talk of updates but they haven't happened yet.

arashi-dev:
Any updates on this? Any help needed? We are currently using https://github.com/MnlPhlp/tauri-plugin-blec package which uses btleplug and we really need this feature to clear peripherals. if any testing or help needed, I am here!

qdot:
Rebased this and brought it in, added noop for linux, will be in v0.12

PR #384, opened 2024-05-11
Use correct size for hci_filter type

sooda:
This is for tag 0.5.5, not the master branch, but github doesn't let me create a PR with the tag as a base. Anyway... My proposed fix for issue 383.

Use 16 bytes for the filter structure in set_socket_filter(), not 14.

struct hci_filter unfortunately is not a packed struct:

https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/lib/hci.h

Hence, it has two bytes of padding at the end for correct stride of the 32-bit members.

This used to not be a problem until Linux kernel commit b2186061d6043 (Bluetooth: hci_sock: Fix not validating setsockopt user input) changed the behaviour of HCI_FILTER sockopt; before, the kernel would have truncated the read size but now the data supplied has to be at least the struct size, or we'd get -EINVAL.

### DIFF
--- a/src/bluez/adapter/mod.rs
+++ b/src/bluez/adapter/mod.rs
@@ -245,7 +245,7 @@ impl ConnectedAdapter {
     }
 
     fn set_socket_filter(&self) -> Result<()> {
-        let mut filter = BytesMut::with_capacity(14);
+        let mut filter = BytesMut::with_capacity(16);
         let type_mask =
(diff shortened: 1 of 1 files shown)


qdot:
That's not a version we're maintaining anymore, so I guess I'll let this hang out but I have nowhere to merge it.

PR #385, opened 2024-05-17
Update platform.rs to add FreeBSD support

nkosi23:
Adding FreeBSD support to platform.rs as it affects the build of a downstream project


qwandor:
Thanks for the PR! You'll need to update the platform-specific dependencies in `Cargo.toml` too. It would also be good to have CI cover this if possible.

qdot (changes requested):
Marking this as changes requested since we'd still need the Cargo.toml updates for platform. CI'ing this may be a challenge.

qdot:
Closing since we've never gotten an update on it and have no way to test.

Issue #387, opened 2024-05-23
On MacOS notifications stream freezes forever when computer locked 


dmtrKovalenko:
With the basic provided example for notification stream

```rust
use btleplug::api::{Central, CharPropFlags, Manager as _, Peripheral, ScanFilter};
use btleplug::platform::Manager;
use futures::stream::StreamExt;
use std::error::Error;
use std::time::Duration;
use tokio::time;
use uuid::Uuid;

/// Only devices whose name contains this string will be tried.
const PERIPHERAL_NAME_MATCH_FILTER: &str = "CO2CICKA Sensor";
/// UUID of the characteristic for which we should subscribe to notifications.
const NOTIFY_CHARACTERISTIC_UUID: Uuid = Uuid::from_u128(0x0000FFE1_0000_1000_8000_00805F9B34FB);

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    pretty_env_logger::init();

    let manager = Manager::new().await?;
    let adapter_list = manager.adapters().await?;
    if adapter_list.is_empty() {
        eprintln!("No Bluetooth adapters found");
    }

    for adapter in adapter_list.iter() {
        println!("Starting scan...");
        adapter
            .start_scan(ScanFilter::default())
            .await
            .expect("Can't scan BLE adapter for connected devices...");
        time::sleep(Duration::from_secs(2)).await;
        let peripherals = adapter.peripherals().await?;

        if peripherals.is_empty() {
            eprintln!("->>> BLE peripheral devices were not found, sorry. Exiting...");
        } else {
            // All peripheral devices in range.
            for peripheral in peripherals.iter() {
                let properties = peripheral.properties().await?;
                let is_connected = peripheral.is_connected().await?;
                let local_name = properties
                    .unwrap()
                    .local_name
                    .unwrap_or(String::from("(peripheral name unknown)"));
                println!(
                    "Peripheral {:?} is connected: {:?}",
                    &local_name, is_connected
                );
                // Check if it's the peripheral we want.
                if local_name.contains(PERIPHERAL_NAME_MATCH_FILTER) {
                    println!("Found matching peripheral {:?}...", &local_name);
                    if !is_connected {
                        // Connect if we aren't already connected.
                        if let Err(err) = peripheral.connect().await {
                            eprintln!("Error connecting to peripheral, skipping: {}", err);
                            continue;
                        }
                    }
                    let is_connected = peripheral.is_connected().await?;
                    println!(
[...]

PR #388, opened 2024-05-24
fix(macos): Break the corebluetooth loop when manager turned off

dmtrKovalenko:
closes https://github.com/deviceplug/btleplug/issues/387

In Core Bluetooth when the device is locked or suspended and it is not applicable for background Bluetooth access the manager will create an event for state change and after will change the manager state to powerOff. Currently, it is not tracked at all which
leads to the forever stuck unresolved futures while the connection to peripheral is still held.

An additional problem I faced that there is no way to manually kill the
event loop of the corebluetooth from outside so the
`CoreBluetoothInternal::drop` is never called because it is always
living in the stalled thread.

In this change, I added an API to access the manager state and exited
the event loop when if the manager turned off.

### DIFF
--- a/src/corebluetooth/adapter.rs
+++ b/src/corebluetooth/adapter.rs
@@ -2,6 +2,7 @@ use super::internal::{run_corebluetooth_thread, CoreBluetoothEvent, CoreBluetoot
 use super::peripheral::{Peripheral, PeripheralId};
 use crate::api::{Central, CentralEvent, ScanFilter};
 use crate::common::adapter_manager::AdapterManager;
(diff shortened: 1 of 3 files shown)


qdot (approved):

qdot:
Ok, well, should've merged this before the corebluetooth API change. Patch looks fine but I need to go back and see what the conflicts are and if this is still valid given the event loop changes made during the API change.

qdot:
We'll be bringing in a fix for this in v0.12, that actually goes through and cancels all futures when we see the poweredoff event.

PR #389, opened 2024-06-01
Use `objc2-core-bluetooth`

madsmtm:
Continuing from https://github.com/deviceplug/btleplug/pull/381, I've created the `objc2-core-bluetooth` crate, so let's use that to avoid having to define the methods ourselves.

Also, we now use `objc2`'s `declare_class!` macro for the custom delegate, this is a lot less error-prone that using `ClassBuilder`. This part was done in two commits to hopefully make the diff easier to read.


qdot (approved):

qdot:
Wow, this is much cleaner than what we had! Thanks!

Issue #334, opened 2023-08-23
Wrong peripheral name on macOS


Massimiliano-solutiontech:
**Describe the bug**
`properties.local_name` doesn't have the correct name because [macOS caches the names](https://developer.apple.com/forums/thread/19381#answers-title)

**Expected behavior**
Have the correct advertised name 

**Actual behavior**
Returning the cached named

**Additional context**
Below the Swift code working on iOS, iPadOS, but I didn't tested it on macOS

https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerdelegate/1518937-centralmanager

```swift
func centralManager(_ central: CBCentralManager, didDiscover peripheral: CBPeripheral, advertisementData: [String: Any], rssi RSSI: NSNumber) {
    guard let name = advertisementData["kCBAdvDataLocalName"] as? String else { return }
}
```

I have experience with Swift and iOS development, if you need help to implement that I'll be happy to do so

Massimiliano-solutiontech:
Any news on that?

gmallios:
I've implemented this and checked if the advertisement data contains the LocalName but no luck. 
Connecting to the device upon discovery seems to solve this issue.

yuyoyuppe:
Having the same issue, the fix by @Massimiliano-solutiontech has helped. In my case `kCBAdvDataLocalName` contained a far more descriptive name.

Specifically, my treadmill is displayed as "LifeSpan-TM-2000" when scanning on Windows with `btleplug`/nrfconnect on Android. However, on macOS 14.5 it's displayed as "ESP32".

I'm not sure what's the most appropriate fix would be, since exposing a new field in a `Peripheral` for all platforms would likely be an overkill, but I guess we can merge names on macOS If both fields are present?

qdot:
I *think* I'll be fine merging them, but I'll have to try them in my own library and see what it does.

yuyoyuppe:
@qdot rebased against `dev`.

qdot:
> *This was generated by AI during triage.*

Closing this issue because current CoreBluetooth name handling preserves the advertisement name and has been corrected since this report.

PR #390, opened 2024-06-06
fix: make local_name in corebluetooth more descriptive

yuyoyuppe:
# Issue description
Corebluetooth stack discovers my BTLE device with a very generic name while having a correct name on Windows and on Android.

# Change description
This PR implements the fix suggested in https://github.com/deviceplug/btleplug/issues/334 and extends the event_driven_discovery example to print peripheral names.

# Concerns
I'm not sure it's the best approach to merge `Advertisement Data[Local Name]` and peripheral names when they're both present, so I'm open for better ideas.

### DIFF
--- a/src/corebluetooth/utils/mod.rs
+++ b/src/corebluetooth/utils/mod.rs
@@ -16,11 +16,20 @@
 // This file may not be copied, modified, or distributed except
 // according to those terms.
 
-use objc2_foundation::NSUUID;
+use std::ffi::CStr;
+
+use objc2_foundation::{NSString, NSUUID};
 use uuid::Uuid;
 
 pub mod core_bluetooth;
 
 pub fn nsuuid_to_uuid(uuid: &NSUUID) -> Uuid {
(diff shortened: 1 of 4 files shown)


yuyoyuppe:
This might break people relying on an exact device name, so would be worth mentioning in Release notes.

src/corebluetooth/utils/mod.rs
```diff
@@ -16,11 +16,22 @@
...
+use objc2_foundation::{NSString, NSUUID};
 use uuid::Uuid;
 
 pub mod core_bluetooth;
 
 pub fn nsuuid_to_uuid(uuid: &NSUUID) -> Uuid {
     uuid.UUIDString().to_string().parse().unwrap()
 }
+
+pub fn nsstring_to_string(nsstring: *const NSString) -> Option<String> {
```
qwandor, line 30: This function should be marked as `unsafe`, as it can cause undefined behaviour if `nsstring` is not a valid pointer.

qwandor (approved):

PR #391, opened 2024-07-15
fix(macos): reading descriptor value

kovapatrik:
Reading out the value of the descriptors was not correctly implemented before. My solution needs to be extended with all the possible standards descriptor UUIDs but it provides a good starting point.

Also fixes a delegate bug (there were incorrect methods assigned to the functions)

### DIFF
--- a/src/corebluetooth/central_delegate.rs
+++ b/src/corebluetooth/central_delegate.rs
@@ -599,7 +599,7 @@ declare_class!(
             }
         }
 
-        #[method(peripheral:didUpdateNotificationStateForCharacteristic:error:)]
+        #[method(peripheral:didWriteValueForCharacteristic:error:)]
         fn delegate_peripheral_didwritevalueforcharacteristic_error(
             &self,
             peripheral: &CBPeripheral,
@@ -622,7 +622,7 @@ declare_class!(
             }
         }
 
-        #[method(peripheral:didWriteValueForCharacteristic:error:)]
+        #[method(peripheral:didUpdateNotificationStateForCharacteristic:error:)]
         fn delegate_peripheral_didupdatenotificationstateforcharacteristic_error(
             &self,
             peripheral: &CBPeripheral,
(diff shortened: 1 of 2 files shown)


src/corebluetooth/central_delegate.rs
```diff
@@ -749,6 +749,21 @@ fn get_characteristic_value(characteristic: &CBCharacteristic) -> Vec<u8> {
     v.unwrap_or_default()
 }
 
+fn get_descriptor_value(descriptor: &CBDescriptor) -> Vec<u8> {
+    trace!("Getting data!");
+    let v = unsafe { descriptor.value() }.map(|value| unsafe {
+        match descriptor.UUID().UUIDString().to_string().as_str() {
+            "2901" => {
```
qwandor, line 756: What's special about "2901" here? Why can't this work for other descriptor UUIDs?

kovapatrik: there are predefined UUIDs for descriptors for which we know the type for (in case of `2901`, it is NSString). I think it's possible to create custom descriptors, but without knowing what the underlying type if for that it's not possible to get the value (AFAIK)

The predefined descriptor UUIDs are in this pdf, in section 3.7:

https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Assigned_Numbers/out/en/Assigned_Numbers.pdf?v=1720988510534

kovapatrik: Forgot to mention the root problem: `descriptor.value()` is a `Retained<AnyObject>`. It must be casted to a type and it will throw an error if you try to cast it to an invalid type. So in this case if I would try to cast the value to for example `NSData` instead of `NSString`, it will throw an error.

qwandor: From what I can see in the Apple documentation, it looks like the value will either be an `NSString`, `NSData`, `NSNumber` or `UInt16`. Can you use `AnyObject::class` to check which type it is, then cast and convert it accordingly? This should be more robust and general than special-casing each UUID.

kovapatrik: I had to trace back to the superclass of the class because in my case I have the same descriptor within 5 different characteristic and in one characteristic the type was `NSTaggedPointerString`, in the other one it was `__NSCFString`. Both are subclasses of `NSString` and can be safely cast to it. This way it's a bit hacky to get the value of `NSNumber`. Do you have any suggestions how should I get the class of a number this way? Also, should we panic if there are no matching class type?

qwandor (approved):

Issue #364, opened 2024-02-01
Unable to scan on windows 10 tablet: "no such interface supported"


ChristianPavilonis:
**Describe the bug**
When calling `adapter.start_scan().await` I get this error: Error { code: HRESULT(0x800004002), message "no such interface supported" }

It only happens on the windows 10 tablet that I'm targeting other windows devices work.

**Expected behavior**
I expect it to start scanning like it does on every other device I test.

**Actual behavior**
The error above.

**Additional context**
The device is a faytech "10.1 Ruggedized Tablet PC"

It runs Windows 10 enterprise LTSC. 

Could it be the fact that it's an LTSC tablet that's is simply so out of date that it wont work with  the `windows` crate?

Might be out of scope for this crate but I'm not sure how to trouble shoot this. 🤔 

I made it panic so I could give you a stack trace.
```rust
thread 'tokio-runtime-worker' panicked at src\ble\discovery.rs:85:53:
called Result::unwrap() on an Err value: Other("Error { code: HRESULT(0x80004002), message: "No such interface supported" }")
stack backtrace:
0: std::panicking::begin_panic_handler
at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\std\src\panicking.rs:645
1: core::panicking::panic_fmt
at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\core\src\panicking.rs:72
2: core::result::unwrap_failed
at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\core\src\result.rs:1653
3: enum2$<core::result::Result<tuple$<>,enum2$btleplug::Error > >::unwrap<tuple$<>,enum2$btleplug::Error >
at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112\library\core\src\result.rs:1077
4: windows_app::ble::discovery::event_based_scan::async_fn$0
at .\src\ble\discovery.rs:85
5: windows_app::main::async_block$0::closure$0::async_block$0
at .\src\main.rs:82
6: tokio::runtime::task::core::impl$6::poll::closure$0<enum2$<windows_app::main::async_block$0::closure$0::async_block_env$0>,alloc::sync::Arctokio::runtime::scheduler::multi_thread::handle::Handle,alloc::alloc::Global >
at C:\Users\fay.cargo\registry\src\index.crates.io-6f17d22bba15001f\tokio-1.35.1\src\runtime\task\core.rs:328
[...]

PR #402, opened 2024-10-20
Supports Windows 10 versions before 2004

wuwbobo2021:
Fixes for this issue: <https://github.com/deviceplug/btleplug/issues/364>.

Thanks to @ChristianPavilonis for providing the clue. He had forked this project and deleted the line that causes the problem.

This fix tries to keep the functionality as is, merely ignoring the returned error if `SetAllowExtendedAdvertisements()` isn't available.

### DIFF
--- a/src/winrtble/ble/watcher.rs
+++ b/src/winrtble/ble/watcher.rs
@@ -51,7 +51,7 @@ impl BLEWatcher {
         self.watcher
             .SetScanningMode(BluetoothLEScanningMode::Active)
             .unwrap();
-        self.watcher.SetAllowExtendedAdvertisements(true)?;
+        let _ = self.watcher.SetAllowExtendedAdvertisements(true);
         let handler: TypedEventHandler<
             BluetoothLEAdvertisementWatcher,
             BluetoothLEAdvertisementReceivedEventArgs,


qdot:
I'd really rather this be a conditional on the OS version rather then just throwing out the error completely, especially because this is such an old version of Win10. Also this needs comments for why it's happening.

qdot (changes requested):
Make this a conditional that parses OS version, because I want to throw an error whenever possible. Also, needs comments about why this is happening since it's done for a rare platform.

wuwbobo2021:
I'm switching to [bluest](https://docs.rs/bluest/0.6.7/bluest/) because the only maintainer of `btleplug` is a sex toy maker from "Nonpolynomial LLC". It's unlikely for me to use this crate anymore.

I has been annoyed by the huge horrible icon in <https://crates.io/crates/buttplug>, which I occasionally discovered before checking the meaning of "butt" in the dictionary. After that I decided to report that crate on crates.io, however the reporting web page doesn't seems to be responding (I make deal with the crates.io bug later).

<https://www.rust-lang.org/policies/code-of-conduct>
<https://github.com/stumpsyn/policies/blob/master/citizen_code_of_conduct.md> (4. Unacceptable Behavior)

PR #407, opened 2024-11-12
Add MTU support across all platforms

robbym:
This PR will include a method to retrieve the negotiated MTU size on each platform.

Info grabbed from: [deviceplug/btleplug#246](https://github.com/deviceplug/btleplug/issues/246)

Windows - [GattSession.MaxPduSize](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattsession.maxpdusize?view=winrt-22621)
Negotiation happens on user's behalf, and a GattSession.MaxPduSizeChanged event is emitted when MTU is updated.

macOS / iOS - [CBPeripheral.maximumWriteValueLength](https://developer.apple.com/documentation/corebluetooth/cbperipheral/1620312-maximumwritevaluelength)
Negotiation happens on user's behalf, have used peripheral.maximumWriteValueLength(for: .withoutResponse) reliably in the past. .withResponse has some issues. [See here.](https://stackoverflow.com/a/68389331)

Linux
Instead of a dedicated MTU property, BlueZ has decided to adopt the newer Bluetooth 5.2 standard. Example here: [Link](https://github.com/OpenBluetoothToolbox/SimpleBLE/blob/614a5af35cdd22ba08318cb776795ecd2da8c389/simpleble/src/backends/linux/PeripheralBase.cpp#L57)

Android - [BluetoothGatt.requestMtu](https://developer.android.com/reference/android/bluetooth/BluetoothGatt#requestMtu(int))
The way I understand it and have used it in the past, you call requestMtu with the largest you can support (e.g. 512), and it will negotiate the min(host.maxMtu, periph.maxMtu) and call onMtuChanged with the new mtu size.

TODO:

- [x] Windows
- [x] Linux
- [ ] Android
- [ ] macOS/iOS

### DIFF
--- a/src/bluez/peripheral.rs
+++ b/src/bluez/peripheral.rs
@@ -138,6 +138,17 @@ impl api::Peripheral for Peripheral {
         self.mac_address
     }
 
+    fn mtu(&self) -> u16 {
+        let services = self.services.lock().unwrap();
+        for (_, service) in services.iter() {
+            for (_, characteristic) in service.characteristics.iter() {
+                return characteristic.info.mtu.unwrap();
+            }
+        }
+        
+        api::DEFAULT_MTU_SIZE
+    }
+
     async fn properties(&self) -> Result<Option<PeripheralProperties>> {
         let device_info = self.device_info().await?;
         Ok(Some(PeripheralProperties {
--- a/src/winrtble/peripheral.rs
+++ b/src/winrtble/peripheral.rs
@@ -46,9 +46,6 @@ use uuid::Uuid;
 use std::sync::Weak;
 use windows::Devices::Bluetooth::{Advertisement::*, BluetoothAddressType};
 
-/// The default MTU size for a peripheral.
-const DEFAULT_MTU_SIZE: u16 = 23;
-
 #[cfg_attr(
     feature = "serde",
     derive(Serialize, Deserialize),
@@ -97,7 +94,7 @@ impl Peripheral {
                 adapter,
                 device: tokio::sync::Mutex::new(None),
                 address,
-                mtu: AtomicU16::new(DEFAULT_MTU_SIZE),
+                mtu: AtomicU16::new(api::DEFAULT_MTU_SIZE),
                 connected: AtomicBool::new(false),
                 ble_services: DashMap::new(),
                 notifications_channel: broadcast_sender,
--- a/src/winrtble/peripheral.rs
+++ b/src/winrtble/peripheral.rs
@@ -46,6 +46,9 @@ use uuid::Uuid;
 use std::sync::Weak;
 use windows::Devices::Bluetooth::{Advertisement::*, BluetoothAddressType};
 
+/// The default MTU size for a peripheral.
+const DEFAULT_MTU_SIZE: u16 = 23;
+
 #[cfg_attr(
     feature = "serde",
     derive(Serialize, Deserialize),
@@ -94,7 +97,7 @@ impl Peripheral {
                 adapter,
                 device: tokio::sync::Mutex::new(None),
                 address,
-                mtu: AtomicU16::new(23),
+                mtu: AtomicU16::new(DEFAULT_MTU_SIZE),
                 connected: AtomicBool::new(false),
                 ble_services: DashMap::new(),
                 notifications_channel: broadcast_sender,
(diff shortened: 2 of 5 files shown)


src/winrtble/peripheral.rs
```diff
@@ -93,6 +94,7 @@ impl Peripheral {
                 adapter,
                 device: tokio::sync::Mutex::new(None),
                 address,
+                mtu: AtomicU16::new(23),
```
qwandor, line 97: Where does this 23 come from?

robbym: It's the default size of the MTU prior to negotiation. It includes the L2CAP header (4 bytes), so usable size would be 19.

robbym: I think I'll add an initial assignment from MaxPduSize prior to the event handler registration.

qwandor: Please add a constant for it, with a comment explaining this.

laxian:
@robbym Hi, have you finish it on bluez?

robbym:
@laxian I just pushed a working bluez implementation. I tested it on one of my BLE devices and it works, but let me know if there is any issues.

@qwandor I added the constant and now moved it to the api namespace so that each implementation can use it. Not really happy with where it currently lives. Maybe move it to common?

src/bluez/peripheral.rs
```diff
@@ -138,6 +138,17 @@ impl api::Peripheral for Peripheral {
         self.mac_address
     }
 
+    fn mtu(&self) -> u16 {
+        let services = self.services.lock().unwrap();
+        for (_, service) in services.iter() {
+            for (_, characteristic) in service.characteristics.iter() {
+                return characteristic.info.mtu.unwrap();
```
qwandor, line 145: Can this be different for different characteristics?

qdot:
Windows parts of this are looking good so far. Would like to get all platforms in before we land it (otherwise I *will* have people on unsupported platforms yelling) but I'm fine to also do a point release whenever we get all the parts in.

qdot:
Converted this to a draft until we get all the platforms in. Just makes my life easier for the little amount of time I actually get to do triage on this project.

qdot:
One more quick update here: I brought us up to bluez-async 0.8.0 in v0.11.7, so mtu() is available now and we can drop that change here (it'll probably just skip the commit on rebase).

src/bluez/peripheral.rs
robbym: In Bluetooth 5.2 and forward, yes. The reason they implemented it as a per-characteristic MTU is because of the enhanced ATT (EATT) bearer feature.

They mention it here: https://github.com/bluez/bluez/issues/199
I mention it here: https://github.com/deviceplug/btleplug/issues/246#issuecomment-1545186841

The majority of devices on the market are going to be using the unenhanced version (single MTU per connection), so I don't understand why bluez didn't add support for that first.

In any case, the implementation was based on how SimpleBle did it: https://github.com/OpenBluetoothToolbox/SimpleBLE/blob/614a5af35cdd22ba08318cb776795ecd2da8c389/simpleble/src/backends/linux/PeripheralBase.cpp#L57

qwandor: Okay, seems reasonable given the API limitations.

qwandor (approved):
BlueZ implementation seems fine.

qdot:
Ok, having to rebase this elsewhere while trying to get the other platforms added in, so:

- Closing this
- Rebasing and fixing conflicts in original commits
- Working out of `pr-407-mtu` to try to bring in all platforms
- Doing a followup to actually negotiate MTUs on all platforms.

WernersGit:
Without a correct mtu() value, applications cannot determine whether large payloads are safe for notifications. In my case I'm streaming PCAP data over GATT notifications and need to know
    whether to fragment at 244 B (BLE 5.0 default ATT_MTU) or 509 B (ATT_MTU 512). I ended up implementing an application-level write probe as a workaround, which is fragile and shouldn't be
    necessary given the feature table says "Retrieve MTU: ✓" for macOS.

    Suggested fix:
    In src/corebluetooth/peripheral.rs, after discover_services() resolves, read maximumWriteValueLengthForType:CBCharacteristicWriteWithoutResponse via objc2 and store the result in
    self.shared.mtu:

    let max_write = msg_send![cb_peripheral, maximumWriteValueLengthForType: 1u64]; // 1 = withoutResponse
    self.shared.mtu.store(max_write as u16 + 3, Ordering::Relaxed); // convert payload length to ATT_MTU

    That single call is the macOS equivalent of what Android already does (correctly ?).

PR #347, opened 2023-11-17
depend on vendored dbus in order to avoid build failure with libdbus-sys on linux aarch64 target

NuLL3rr0r:
Cross-compiling for Linux AArch64 fails due to libdbus-sys native library dependency. As stated in the [dbus-rs compilation guide](https://github.com/diwic/dbus-rs/blob/master/libdbus-sys/cross_compile.md) there are three ways to resolve this issue. The simplest one would be:

```
dbus = {version = "0.9.7", features = ["vendored"]}
```

Otherwise, when cross-compiling for AArch64 from a x86-64 Linux host, the user gets:

```
   Compiling proc-macro2 v1.0.69
   Compiling unicode-ident v1.0.12
   Compiling autocfg v1.1.0
   Compiling libc v0.2.150
   Compiling futures-core v0.3.29
   Compiling futures-sink v0.3.29
   Compiling pin-project-lite v0.2.13
   Compiling futures-channel v0.3.29
   Compiling futures-task v0.3.29
   Compiling futures-util v0.3.29
   Compiling pin-utils v0.1.0
   Compiling futures-io v0.3.29
   Compiling pkg-config v0.3.27
   Compiling memchr v2.6.4
   Compiling serde v1.0.192
   Compiling thiserror v1.0.50
   Compiling parking_lot_core v0.9.9
   Compiling async-trait v0.1.74
   Compiling either v1.9.0
   Compiling scopeguard v1.2.0
   Compiling xml-rs v0.8.19
   Compiling cfg-if v1.0.0
   Compiling smallvec v1.11.2
   Compiling log v0.4.20
   Compiling bytes v1.5.0
   Compiling itertools v0.10.5
   Compiling slab v0.4.9
   Compiling lock_api v0.4.11
   Compiling bitflags v2.4.1
   Compiling hashbrown v0.14.2
   Compiling once_cell v1.18.0
   Compiling uuid v1.5.0
   Compiling static_assertions v1.1.0
   Compiling libdbus-sys v0.2.5
   Compiling quote v1.0.33
error: failed to run custom build command for `libdbus-sys v0.2.5`

Caused by:
  process didn't exit successfully: `/target/release/build/libdbus-sys-a5bd944697d4bc4a/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=build_vendored.rs
  cargo:rerun-if-env-changed=DBUS_1_NO_PKG_CONFIG
[...]

PR #411, opened 2024-12-08
Add support for building vendored dbus

twistedfall:
Hi! This PR adds a new feature `dbus-vendored` which enables the build against vendored `dbus` as opposed to using system `.so`. I needed this to make the successful cross-platform build. It was previously discussed in the PR https://github.com/deviceplug/btleplug/pull/347 and it looks like the agreement was to add an optional feature for that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ path = "src/lib.rs"
 
 [features]
 serde = ["uuid/serde", "serde_cr", "serde_bytes"]
+dbus-vendored = ["dbus/vendored"]
 
 [dependencies]
 async-trait = "0.1.83"


qwandor:
Why not add this to your own binary's `Cargo.toml` rather than `btleplug`?

twistedfall:
@qwandor That would work, but that would mean my crate would need to depend on dbus too (which it currently doesn't). And I will need to replicate all of the conditional platform support logic to make sure I'm not including dbus for macOS build for example.

qwandor:
I don't really think it belongs here. If someone wants to build their binary with a vendored copy of libdbus they can add the dependencies directly to their binary.

qdot:
'k, closing.

PR #414, opened 2024-12-11
Fix #227, use BluetoothCacheMode::Uncached when discovering services for device

### LINKED ISSUE #227: discover_services() can return different (and less) results when called multiple times on windows.
**Describe the bug**
1. On windows 10, connect to a device and call discover_services()
2. Without touching any characteristics, call discover_services() again fairly soon after on the same device

**Expected behavior**
Nothing should change.

**Actual behavior**
At least on one device (Satisfyer Curvy 2+), the first call enumerates everything, then the second call loses a bunch of characteristics.

Bradley-Cooley:


Bradley-Cooley:
Hi there, I encountered the issue described in #227, modifying the cache mode solved it for me on Windows 11.

qdot:
I'd been avoiding this because there'd been talk in the past of windows taking a *very* long time to connect when using uncached services (for instance, https://github.com/hbldh/bleak/pull/695). That said, apparently the cache is only used when a device is paired and is cleared on unpair? Also that was 2 years ago and there's been many changes to the windows bluetooth stack since. We'll give this a shot and see what happens.

qdot (approved):

PR #415, opened 2024-12-18
Deserializing BDAddr from non-borrowed string without clone

ekuinox:
There was a problem with deserialization when using [toml](https://crates.io/crates/toml) crates, so I fixed it.

I often use toml files to configure my own software, but the following deserialization using toml was failing.

```rust
#[derive(Deserialize, PartialEq, Copy, Clone, Debug)]
struct Data {
    addr: BDAddr,
}

toml::from_str(r#"addr = "ff00ff00ff00""#) // error: A colon seperated Bluetooth address, like `00:11:22:33:44:55`
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,3 +81,4 @@ rand = "0.8.5"
 pretty_env_logger = "0.5.0"
 tokio = { version = "1.40.0", features = ["macros", "rt", "rt-multi-thread"] }
 serde_json = "1.0.128"
+toml = "0.8.19"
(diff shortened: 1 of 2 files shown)


qwandor (approved):

ekuinox:
It seems that it was already supported in the dev branch. (70ee65cc)
This was handled using `Cow`, but I don't see a problem with returning `BDAddr` directly in the Visitor. I think it would eliminate the need to allocate `Cow`.

qwandor (approved):

qwandor:
Thanks!

Issue #403, opened 2024-10-26
write_value Error { code: HRESULT(0x80000013), message: "该对象已关闭。" }


zhuyu4839:
**Describe the bug**
The function at file `winrtble\ble\characteristic.rs`
`pub async fn write_value(&self, data: &[u8], write_type: WriteType) -> Result<()> {
    ....
    let result = operation.await?;
    ....
}`
return an error:
{ code: HRESULT(0x80000013), message: "该对象已关闭。" }

**Expected behavior**

**Actual behavior**

**Additional context**
Windows info:
OS 名称:          Microsoft Windows 10 企业版 LTSC
OS 版本:          10.0.17763 Build 17763
![image](https://github.com/user-attachments/assets/97916a94-ea6f-4d87-b79a-6696ff9a3ab0)

PR #425, opened 2025-04-11
Update windows to 0.61

jesses2025smith:
1. update `windows` to 0.61 and fix;
2. add `anyhow` and update `rand` for example;
3. add [`PoisonError`] to [`Error`] for replace `unwrap` to `map_err`;


jesses2025smith:
fixed #403

qdot:
In windows-rs 0.61, how does .get() work? Is that the blocking form, since it isn't awaited now?

qdot:
Ok, yup, answered my own question by looking up the relevant windows-rs PR: https://github.com/microsoft/windows-rs/pull/3213

.get() is blocking. This can't come in, we need to use the IntoFuture() casts and await as we currently do.

qdot:
Ok, I did the fix myself 'cause it was simple and I'd like to get windows 0.61 in.

Issue #432, opened 2025-05-18
Get Adapter Mac Address


DanielHabenicht:
**Feature Description**

To know which adapter is selected / select the right adapter it would be great to be able to access its mac address. 
As [`adapter_info` should not be relied on](https://github.com/deviceplug/btleplug/blob/c227cc1adf885e143217f274758968afc02490fb/src/api/mod.rs#L372-L376). 

Example: 
```rust
impl Central for Adapter {
    async fn adapter_mac(&self) -> Result<BDAddr> {
        // Get actual adapter address...
        Ok(BDAddr::from("B8:27:EB:7C:51:30"))
    }
}
```

qdot:
Welp, yet another "not gonna work on all our platforms" issue. Android just gives everyone 02:00:00:00:00:00 for security reasons (this is a fingerprintable address after all). I'm guessing iOS just straight up bans access to it, wouldn't be surprised if macOS is the same. So we've gotta figure out what that's going to look like on the two platforms you didn't provide solutions for before we can do this.

DanielHabenicht:
Sorry ;D

We could document it in the Readme and just give back `02:00:00:00:00:00` or `00:00:00:00:00:00` instead of the `todo!`. 
I mean if it is banned by the platform itself there is probably no a workaround available and if it is it would (hopefully) be patched...

Fristi:
If you got control over the BLE service itself, you can add a service yourself which you can query for it's address

qdot:
> *This was generated by AI during triage.*

Closing this issue because the requested adapter-address API is implemented; macOS correctly reports that the platform does not expose a hardware address.

PR #433, opened 2025-05-18
add adapter mac address

DanielHabenicht:
#432

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ objc2-core-bluetooth = { version = "0.2.2", default-features = false, features =
 ] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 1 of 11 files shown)


DanielHabenicht:
I don't have a mac, nor the time to setup a android development environment. So hopefully thats sufficient for a first implementation?

qdot:
Blocked on discussion in #432

gabevenberg:
Even with just the impls currently in place, this would be very helpful for those of us developing linux specific stuff. Id love it if it was possible to expose the entire contents of bluezs `AdapterInfo` struct.

qdot:
Instead of asking for updates on a 16 month old PR, ended up doing my own impl of this, will be coming in as part of 0.13.0. Same basic idea, just returns Result<Option<BTAddr>> to cover all the possible outcomes. Exposed as `Central::adapter_address()`.

Issue #259, opened 2022-06-20
can not find device already connect to computer


dongjian:
**Describe the bug**
  
I have a keyboard that communicates through GATT. When he is not linked to the computer, it can be searched normally, but when it is linked to the computer, it cannot be found. But I can find the device normally (whether it is already connected to the computer or not) using "Bluetooth LE Explorer".

**Expected behavior**
device can be find normal

**Actual behavior**
can not  find device when it connect to computer

I test both on Windows 10 and Mac 11.6.4

dongjian:
My test code

    let manager = Manager::new().await.unwrap();
    let adapter_list = manager.adapters().await.unwrap();

    println!("adapter len  === {}", adapter_list.len());
    let central = manager
        .adapters()
        .await
        .expect("Unable to fetch adapter list.")
        .into_iter()
        .nth(0)
        .expect("Unable to find adapters.");
    let mut events = central.events().await.unwrap();
    central.start_scan(ScanFilter::default()).await.unwrap();
    loop {
        tokio::time::sleep(Duration::from_secs(2)).await;
        for p in central.peripherals().await.unwrap() {
            println!("{:?}",p.properties().await.unwrap().unwrap().local_name);
            if p.properties()
                .await
                .unwrap()
                .unwrap()
                .local_name
                .iter()
                .any(|name| name.contains("RY KB3 5.0"))
            {
                println!("FFFFIIIIINNNNNNNDDDDDDDDD");
            }
        }
    }

qdot:
Does the keyboard still work as an actual keyboard? Is it bonded to your system? If so, this may be the operating system picking it up in the HID manager before anything else can get to it.

dongjian:
> Does the keyboard still work as an actual keyboard? Is it bonded to your system? If so, this may be the operating system picking it up in the HID manager before anything else can get to it.

yes.keyboard work as an actual keyboard. but I can find the device use "Blue Tooth Le Explorer" OR in Chrome use the code below
navigator.bluetooth.requestDevice({
    acceptAllDevices: true,
  })
  .then(device => { /* … */ })
  .catch(error => { console.error(error); });

GPeye:
I'm also getting this issue on win10. If the device is already paired/connected, enumerating the peripherals on an adapter will only yield BT devices that are available for pairing but not yet connected. It seems silly to have to remove/unpair the device and then put it back in pairing mode and let btleplug do a scan, find the device and connect, instead of just being able to find an existing connection.

[...]

PR #437, opened 2025-05-30
Retrieve connected devices

challiwill:
https://github.com/deviceplug/btleplug/issues/259

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ objc2-core-bluetooth = { version = "0.2.2", default-features = false, features =
 ] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows = { version = "0.61", features = ["Devices_Bluetooth", "Devices_Bluetooth_GenericAttributeProfile", "Devices_Bluetooth_Advertisement", "Devices_Radios", "Foundation_Collections", "Foundation", "Storage_Streams"] }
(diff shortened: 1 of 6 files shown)


pyang2045:
@challiwill It appears this PR doesn’t have a reviewer yet

challiwill:
@pyang2045 That's correct. I was hoping to get feedback from @qdot following the conversation in issue https://github.com/deviceplug/btleplug/issues/259, but I figure we've all been busy with other commitments.

szymonlesisz:
im not sure if retrieved connected peripherals should be treated as discovered.

i've played around this for a little bit and from my experience on macos if device is connected to another app you cant really use it, like for example disconnect it and im pretty sure i had some issues with writing or reading from it.
i guess this is the reason why corebluetooth is hiding it from the scan results.

with your changes it would act as it is discovered, connected and ready to go.

similar case is with [retrieving peripherals by identifiers](https://developer.apple.com/documentation/corebluetooth/cbcentralmanager/retrieveperipherals(withidentifiers:)/) 
if device is known (paired) but it is not advertising it will also not be visible in scan results.

instead i'm proposing less platform-specific approach and create [Adapter.retrieve_peripherals()](https://github.com/deviceplug/btleplug/pull/448) with options (by services, by identifiers, ...by other_future_params)

what dou you think? @qdot? @qwandor?

qdot:
Bringing a version of this in via #448, I do like that API surface better.

PR #441, opened 2025-07-21
Fix: corebluetooth write without response

szymonlesisz:
Hi,

I've run into issue with `WriteType::WithoutResponse` on macos.
im trying to send ~2MB of data in shortest time possible however corebluetooth keeps loosing the packects shortly after the start.
```
When you use CBPeripheral.writeValue(_:for:type:) with .withoutResponse, CoreBluetooth:
- Queues the writes internally
- Starts transmitting them during connection intervals.
- Does not notify when it’s safe to send more.
- Silently drops writes if the queue is overwhelmed

```

while looking up for solutions in other bluetooth libs i came across [this technical document](https://developer.apple.com/library/archive/qa/qa1931/_index.html) and [this guide](https://punchthrough.com/core-bluetooth-guide/)

i belive the ultimate fix would be to use `peripheral.peripheralIsReadyToSendWriteWithoutResponse` as described in the guide but i dont see a easy way how to implement this method (missing CBPeripheralDelegate reference)

so i decided to use attempts loop and check every 15ms (`Interval Min` stated in the document) if it is safe to send another packet. the loop ends if true or after 100 retries (1500ms < `connSupervisionTimeout` stated in the document)

The only problem is that `canSendWriteWithoutResponse` was introduced in macos 11.0 (Big Sur) but it behaves unreliably  until 11.2 - returning false until the first write is performed. That is why i've added `can_send_without_response_supported` flag which is set from the `NSProcessInfo`.

if you dont care about older OS releases (Big Sur was released in november 2020) i can drop the the second commit

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ objc2-foundation = { version = "0.2.2", default-features = false, features = [
     "NSString",
     "NSUUID",
     "NSValue",
+    "NSProcessInfo",
 ] }
 objc2-core-bluetooth = { version = "0.2.2", default-features = false, features = [
     "std",
--- a/src/corebluetooth/internal.rs
+++ b/src/corebluetooth/internal.rs
@@ -38,6 +38,7 @@ use std::{
     fmt::{self, Debug, Formatter},
     ops::Deref,
     thread,
+    time::Duration,
 };
 use tokio::runtime;
 use uuid::Uuid;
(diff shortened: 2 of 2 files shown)


szymonlesisz:
i've noticed this [pull request](https://github.com/deviceplug/btleplug/pull/337) where is also OS version/support check added. 

so im wondering maybe should i create CoreBluetoothInternal.support field? for example:
```
struct CoreBluetoothSupport {
    CanSendWithoutResponse: bool,
    ManagerAuthorization: bool,
}
```

qdot:
Thanks for the PR! This is definitely an issue, but I think there's a way to implement using the delegate, as I'd really worry about blocking the corebluetooth thread with waits. We're gonna go with that, should be in for v0.12. We've got a hardware peripheral test system coming in too, and I've added a WriteWithoutResponse burst test to make sure this works.

prusnak:
> We're gonna go with that, should be in for v0.12. We've got a hardware peripheral test system coming in too, and I've added a WriteWithoutResponse burst test to make sure this works.

Has this issue been fixed one way or another in 0.12 or 0.13 ?

qdot:
@prusnak Should've come in around 0.12 I think.

PR #445, opened 2025-10-21
add service uuids to notifications

MnlPhlp:
I want to support multiple characteristics with the same UUID in different services in [tauri-plugin-blec](https://github.com/MnlPhlp/tauri-plugin-blec).
In order to support this for subscribes, the ValueNotification has to specifiy what service caused it.
I added the values for all the integrations except Android, because I don't really know how to work with the jni code.

I can look at the android version, but I would prefer if someone with more experience in this area could do that.

The problem is I need a new release of this library with the added fields in order to release my crate with the new dependency.
If i checked correctly, there is no way to publish to crates.io with my github fork as dependency

### DIFF
--- a/src/bluez/peripheral.rs
+++ b/src/bluez/peripheral.rs
@@ -298,11 +298,10 @@ fn value_notification(
             event: CharacteristicEvent::Value { value },
         } if id.service().device() == *device_id => {
             let services = services.lock().unwrap();
-            let info = find_characteristic_by_id(&services, id.clone())?;
-            let service = services.iter().find(|(_, s)| s.info.id == id.service())?.0;
+            let (charac, service) = find_characteristic_by_id(&services, id.clone())?;
             Some(ValueNotification {
-                uuid: info.uuid,
-                service_uuid: *service,
+                uuid: charac.uuid,
+                service_uuid: service.uuid,
                 value,
             })
         }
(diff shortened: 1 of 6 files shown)


src/bluez/peripheral.rs
```diff
@@ -298,8 +298,13 @@ fn value_notification(
             event: CharacteristicEvent::Value { value },
         } if id.service().device() == *device_id => {
             let services = services.lock().unwrap();
-            let uuid = find_characteristic_by_id(&services, id)?.uuid;
-            Some(ValueNotification { uuid, value })
+            let info = find_characteristic_by_id(&services, id.clone())?;
+            let service = services.iter().find(|(_, s)| s.info.id == id.service())?.0;
```
qwandor: Rather than looping through `services` a second time, how about making `find_characteristic_by_id` return the `ServiceInfo` as well as the characteristic?

MnlPhlp: I changed it to return a tuple with characteristic and service info.

qwandor (approved):

MnlPhlp:
Are there any plans on when you will do the next release?
Sadly I can't do a new release of the tauri plugin while depending on the git branch.
Cargo  requires a published version for all dependencies to publish my crate.

Or do you know any other way I could publish an update?

PR #446, opened 2025-10-22
fix: bluez device disconnection

szymonlesisz:
when device is upaired from the system settings while connected `CentralEvent::DeviceDisconnected` is never emitted.
it fails on line:
```
let device = session.get_device_info(&id).await.ok()?;
```
with error:

`Err(DbusError(D-Bus error: Method "GetAll" with signature "s" on interface "org.freedesktop.DBus.Properties" doesn't exist`

Screenshot from 2025-10-22 13-58-48

### DIFF
--- a/src/bluez/adapter.rs
+++ b/src/bluez/adapter.rs
@@ -143,11 +143,10 @@ async fn central_event(
                 Some(CentralEvent::DeviceDiscovered(device.id.into()))
             }
             DeviceEvent::Connected { connected } => {
-                let device = session.get_device_info(&id).await.ok()?;
                 if connected {
-                    Some(CentralEvent::DeviceConnected(device.id.into()))
+                    Some(CentralEvent::DeviceConnected(id.into()))
                 } else {
-                    Some(CentralEvent::DeviceDisconnected(device.id.into()))
+                    Some(CentralEvent::DeviceDisconnected(id.into()))
                 }
             }
             DeviceEvent::Rssi { rssi: _ } => {


src/bluez/adapter.rs
```diff
@@ -143,11 +143,15 @@ async fn central_event(
...
-                    Some(CentralEvent::DeviceConnected(device.id.into()))
+                // device may be unknown if unpaired manually from the system
+                let device = session.get_device_info(&id).await;
+                if device.is_ok() && connected {
+                    Some(CentralEvent::DeviceConnected(id.into()))
+                } else if !connected {
+                    Some(CentralEvent::DeviceDisconnected(id.into()))
                 } else {
-                    Some(CentralEvent::DeviceDisconnected(device.id.into()))
+                    // unexpected state. connected without device_info
```
qwandor: Does this ever happen? If not, could we could skip the `get_device_info` call entirely?

szymonlesisz: yes, i was thinking about it but wasnt sure if skipping `get_device_info` affects some inner state of bluez_async or local, but seems its safe to skip it in this case :+1: 

btw, this code is done in each other case (calling `get_device_info` just to use its id) - so maybe there is some other reason for this

rebased, forcepushed

qwandor: It shouldn't affect any state, so I'm not sure why that's ended up here. Maybe we were using something else from it in the past?

qwandor (approved):

Issue #434, opened 2025-05-20
Peripheral Discovery Name Behavior Is Inconsistent


tternes:
PR #390 introduced a change to how peripheral names are formatted to discovery observers on platforms using Core Bluetooth. Unfortunately, this change introduces a non-standard format for the name, which consequently introduces inconsistency within the btleplug library. I'd like to explore how to address this inconsistency and ensure a better experience for end users across platforms.

**Expected behavior**
Wherever possible, btleplug should behave consistently across platforms. When a peripheral is discovered, the name should be predicable as much as possible, regardless of whether the discovery is being performed on Windows, macOS, or Linux. In cases where that isn't possible (due to underlying limitations in the host BLE stack, for instance), the behavior should match what is provided or documented by the platform - unless otherwise justified.

On Apple platforms that provide the Core Bluetooth framework, the [name](https://developer.apple.com/documentation/corebluetooth/cbperipheral/name) property is well-documented and the behavior is predictable:

> A peripheral may have two different name types: one that the device advertises and another that the device publishes in its database as its Bluetooth low energy Generic Access Profile (GAP) device name. If a peripheral has both types of names, this property returns its GAP device name.

During discovery, the `local_name` property of a peripheral should be consistent, and its underlying source should be documented. Ideally, the library would be capable of providing both the broadcast name, as well as the GAP name, making it the responsibility of an integrating application to determine what end users see (or how the applications behaves on behalf of users).

**Actual behavior**

However, the current implementation of btleplug changes this behavior, instead doing extra work to format a name that presents both of these values. While this might be useful during development, it is problematic in production scenarios where these names are shown directly to end users.

The specifics of our current situation are related to a device name change after a firmware update. Our application can deal with receiving an old name (which is what Core Bluetooth reports as the `peripheral.name`) as well as the new name (which is inside the advertising data). Unfortunately, neither of those is provided to the discovery event. Instead, we receive a string formatted like this:

`Old Name [New Name]`

**Additional context**
The current implementation [was introduced](https://github.com/deviceplug/btleplug/commit/a1f8644455de65f2b2ef949ff5a981c31e3fad55#diff-4ac0409716d82c57637817e8c82bf25e62506fd9de450c9bb68090346f2d3a95R542) to address name inconsistency with a particular device. I would suggest that this issue was more likely an implementation detail - or possibly a bug - within that specific peripheral.

[...]

PR #447, opened 2025-10-22
fix: corebluetooth peripheral name

szymonlesisz:
as [well described here](https://github.com/deviceplug/btleplug/issues/434) peripheral `local_name` and `advertisement_name`  are two different things and it is justified to keep them separated as situation that local_name is not equal advertisement_name could happen in all platforms and can be implemented later as followup.

currently it is quite unfortunate that the API changes the peripheral local_name (GAP) by joining those two values

changes:
1. added `Peripheral.advertisement_name` - default None
2. implement advertisement_name in corebluetooth

fix: https://github.com/deviceplug/btleplug/issues/434

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -179,8 +179,10 @@ pub struct PeripheralProperties {
     pub address: BDAddr,
     /// The type of address (either random or public)
     pub address_type: Option<AddressType>,
-    /// The local name. This is generally a human-readable string that identifies the type of device.
+    /// The GAP local name. This is generally a human-readable string that identifies the type of device.
     pub local_name: Option<String>,
+    /// The advertisement name. May be different than local_name.
+    pub advertisement_name: Option<String>,
     /// The transmission power level for the device
     pub tx_power_level: Option<i16>,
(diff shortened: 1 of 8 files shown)


tternes:
Thank you for working on this! I should be able to do some testing this week and share feedback.

tternes:
I finally had a chance to test this in our setup. It works exactly as expected. Really nice solution.

Thanks @szymonlesisz!

qdot (approved):
Ok, this does seem like the correct thing to do, and probably should've been done from the start but we're still dealing with issues we inherited from forking 7+ years ago heh.

Main issue is going to be adapting it to the rest of the platforms so we're consistent, but I'll handle that.

tternes:
> Ok, this does seem like the correct thing to do, and probably should've been done from the start but we're still dealing with issues we inherited from forking 7+ years ago heh.
> 
> Main issue is going to be adapting it to the rest of the platforms so we're consistent, but I'll handle that.

I'm happy to help with testing when you have a PR ready.

qdot:
@tternes It's actually sitting on dev now because I wasn't watching what I was doing heh. Feel free to give it a shot.

PR #448, opened 2025-10-23
feat: add Adapter `retrieve_peripherals` method

szymonlesisz:
This is empty prerequisite for:
- retrieve connected peripherals (by services uuid) https://github.com/deviceplug/btleplug/pull/437
- retrieve known peripherals (by device identifiers) - i will add the implementation for corebluetooth, windows and linux in following pull requests.

List of platforms where the method could be implemented

|          | paired/connected visible | retrieve by identifier | retrieve by services |
|----------|--------------------------|------------------------|----------------------|
| macOS    | NO                       | YES                    | YES                  |
| Linux    | YES                      | YES (default)          | YES (default)        |
| Windows  | NO                       | YES                    | YES                  |
| Android  | ?                        | ?                      | ?                    |

im not familiar with `droidplug` but im sure it is also possible to implement if needed.

### DIFF
--- a/src/bluez/adapter.rs
+++ b/src/bluez/adapter.rs
@@ -1,5 +1,5 @@
 use super::peripheral::{Peripheral, PeripheralId};
-use crate::api::{Central, CentralEvent, CentralState, ScanFilter};
+use crate::api::{Central, CentralEvent, CentralState, RetrievePeripheralsOptions, ScanFilter};
 use crate::{Error, Result};
 use async_trait::async_trait;
 use bluez_async::{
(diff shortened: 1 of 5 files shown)


src/bluez/adapter.rs
```diff
@@ -91,6 +91,13 @@ impl Central for Adapter {
             .collect())
     }
 
+    async fn retrieve_peripherals(
+        &self,
+        options: RetrievePeripheralsOptions,
+    ) -> Result<Vec<Peripheral>> {
+        Err(Error::NotSupported("Not implemented".to_string()))
```
szymonlesisz, line 98: maybe instead of throwing i should return an empty `vec` ? and only log/warn about missing implementation?

PR #449, opened 2025-12-03
Android: Initialize jni-utils

tternes:
I'm in the process of integrating this library into an Android application and am hoping to contribute a handful of improvements to help address some of the challenges folks face when getting this to work.

This first change is to push initialization of the `jni-utils-rs` crate into this library directly. This is a critical step to getting callbacks from the Java code to wake up Rust futures, but is not consistently shown in the examples I've tracked down. Some examples establish peer dependencies on `btleplug` and `jni-utils-rs`, but those need to be carefully version-matched, causing fragility (and unfortunately in my testing, silent bugs that are hard to run down).

Since `btleplug` directly depends on this crate, the initialization here made sense. I don't love the `unwrap()` and am very open to ideas about how to better surface these types of early initialization errors to the application.

### DIFF
--- a/src/droidplug/jni/mod.rs
+++ b/src/droidplug/jni/mod.rs
@@ -9,6 +9,8 @@ static GLOBAL_JVM: OnceCell<JavaVM> = OnceCell::new();
 
 pub fn init(env: &JNIEnv) -> crate::Result<()> {
     if let Ok(()) = GLOBAL_JVM.set(env.get_java_vm()?) {
+        jni_utils::init(&env).unwrap();
+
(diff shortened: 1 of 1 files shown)


qdot:
Good news! No one else uses jni_utils but us (it was created for this project), so I've just folded it into droidplug for v0.12. It means we won't need this step or even to have to worry at all about the external repo/dependency anymore.

Thanks for the PR at least. :)

tternes:
Awesome - I have at least one bug fix there I need to push upstream. I'll wait on that until you've landed that library here. Thanks!

PR #453, opened 2026-01-22
Handle peripheral:didModifyServices on corebluetooth

### LINKED ISSUE #343: CoreBluetooth should do service discovery like other platforms
Right now, CoreBluetooth only does service discovery on connect. The discoverServices() method is a no-op. This was done early on just to get macOS working then was never fixed.

kragacles:
When the corebluetooth subsystem emits the didModifyServices event the peripheral services become unusable until service discovery is performed again: https://developer.apple.com/documentation/corebluetooth/cbperipheraldelegate/peripheral(_:didmodifyservices:)?language=objc

This PR adds awareness of the didModifyServices event up through to the API layer so that service discovery may be performed again by the user. It also clears discovered services on the corebluetooth platform. While technically only the modified services are unusable after this event, practical testing showed that the peripheral in general is not functional until a full service discovery is performed again.

While adding this support I realized issue: #343 made it impossible to re-discover services without reconnecting to the device. So, this PR also resolves #343 by splitting service discovery off of connection on corebluetooth and moving it to the standalone discover_services() function.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -319,6 +319,8 @@ pub enum CentralEvent {
     DeviceUpdated(PeripheralId),
     DeviceConnected(PeripheralId),
     DeviceDisconnected(PeripheralId),
+    /// Only emitted on the corebluetooth subsystem
+    DeviceServicesModified(PeripheralId),
     /// Emitted when a Manufacturer Data advertisement has been received from a device
     ManufacturerDataAdvertisement {
         id: PeripheralId,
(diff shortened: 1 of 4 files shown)


qdot:
Thanks! This'll come in in v0.12, and I'll make a callout that it may break expectations on macOS, but it does put the API in line with the other platforms so I'm fine with that.

qdot (approved):
There's a few cleanup things we can do here (don't need clone, etc), but I'll bring this in and handle those since ti's been 6 weeks since this was submitted.

PR #362, opened 2024-01-09
Ignore failed descriptor reads on Windows

DavidZemon:
This brings the Windows implementation in line with Linux

Fixes #361

PR #456, opened 2026-03-18
fix(corebluetooth): send descriptor discovery event even on error

mikatammi:
On macOS, `peripheral:didDiscoverDescriptorsForCharacteristic:error:` only sent the
`DiscoveredCharacteristicDescriptors` event when `error` was nil. If CoreBluetooth reported
an error for any characteristic's descriptors (e.g. "The specified UUID is not allowed for
this operation"), the event was never emitted and `discover_services()` would hang
indefinitely waiting for all characteristics to be marked as discovered.

This is the macOS/CoreBluetooth equivalent of the Windows fix in #362 — descriptor discovery
errors should not block service discovery from completing.

The fix moves the event send and UUID extraction outside the `if error.is_none()` block, so
`DiscoveredCharacteristicDescriptors` always fires. On error, an empty descriptor map is
sent and a warning is logged, which unblocks the discovery future.

Tested with a Shelly Plug SG3 device whose FFF6 service characteristic triggers "The
specified UUID is not allowed for this operation" during descriptor discovery.

### DIFF
--- a/src/corebluetooth/central_delegate.rs
+++ b/src/corebluetooth/central_delegate.rs
@@ -20,7 +20,7 @@ use super::utils::nsstring_to_string;
 use super::utils::{core_bluetooth::cbuuid_to_uuid, nsuuid_to_uuid};
 use futures::channel::mpsc::Sender;
 use futures::sink::SinkExt;
-use log::{error, trace};
+use log::{error, trace, warn};
 use objc2::runtime::{AnyObject, ProtocolObject};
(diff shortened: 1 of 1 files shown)


mikatammi:
The clippy issues probably were already there before my PR? Should we fix them first?

qdot:
Yup, I need to do a ton of clippy cleanup and hadn't been paying attention to actions, your PR is fine. I'll try to review soon.

qdot:
Coming in as part of 0.13.0

PR #461, opened 2026-04-14
fix: Preserve ScanResponse on Windows and resolve cross-platform N/A device names

Yongle-Fu:
This PR resolves 3 critical and long-standing behavior inconsistencies across the Windows and macOS backends, strictly enforcing lifecycle contracts and BLE specifications.

### 1. Common: Fix accidental eviction of connected peripherals
**Problem:** `AdapterManager.clear_peripherals()` was previously unconditionally emptying the internal device map. This meant that whenever scanning was stopped, any active/connected peripherals were instantly silently evicted, killing the connection.
**Fix:** Introduced `.retain()` to filter out connections checking `is_connected()` sync-safe loop. CoreBluetooth/BlueZ semantics are now properly enforced across all platforms. 

### 2. WinRT: Prevent Service-UUID filters from silently dropping ScanResponse
**Problem:** Windows `ScanResponse` packets contain vital identity information (like Name) but often omit the UUIDs found in the primary advertisement. Because the WinRT `watcher` applied the UUID filter strictly to the contents of *every single packet*, all `ScanResponse` packets were silently dropped, resulting in "N/A" device names.
**Fix:** Implemented a lightweight, tightly-scoped `BluetoothAddress` cache within the WinRT `watcher`. If a device's initial advertisement passes the UUID filter, its address is cached, granting safe passage for its subsequent UUIDless `ScanResponse`.

### 3. Cross-platform: Enforce COMPLETE_LOCAL_NAME priority
**macOS (CoreBluetooth):** 
- Inverted priority mapping: `advertisement_name` (Scan Response / Complete Name) is now strongly preferred over the internal `peripheral.name()` GAP cache (which operates as a truncated short name).
- Added length-aware safe-guards inside `update_name()` to prevent subsequent incomplete packets from wiping out a previously parsed long name.

**Windows (WinRT):**
- Parsed `SHORT_LOCAL_NAME (0x08)` and `COMPLETE_LOCAL_NAME (0x09)` directly from `DataSections`.
- Enforced standard BLE precedence where Complete Name overrides Short Name across individual and fragmented packet arrivals.

### DIFF
--- a/src/common/adapter_manager.rs
+++ b/src/common/adapter_manager.rs
@@ -67,7 +67,15 @@ where
     }
 
     pub fn clear_peripherals(&self) {
-        self.peripherals.clear();
+        // Retain peripherals that are currently connected.
+        // This ensures that stopping a scan or clearing discovered devices
+        // does not evict active connections — matching CoreBluetooth/BlueZ behavior.
+        self.peripherals.retain(|_id, peripheral| {
+            // is_connected() is async in the trait, but all platform implementations
+            // use an AtomicBool internally, so block_on is safe here.
+            // We use a short-lived runtime to avoid panics if already in an async context.
+            futures::executor::block_on(peripheral.is_connected()).unwrap_or(false)
+        });
     }
 
     pub fn peripherals(&self) -> Vec<PeripheralType> {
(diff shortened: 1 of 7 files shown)


qdot:
Coming in as part of 0.13.0, with some changes (removed no-op changes in Android to reduce conflicts, fixed up possible blocking dashmap retain in clear_peripherals)

PR #465, opened 2026-07-15
feat(winrtble): implement add_peripheral (connect by address)

tanarchytan:
## What

Implement the WinRT Central::add_peripheral, which currently returns Err(NotSupported). It creates a Peripheral from the given PeripheralId the same way the scanner registers a discovered device, and returns an already-known entry if present. Also adds the symmetric From<PeripheralId> for BDAddr.

## Why

add_peripheral was added to the Central trait in 0.12.0 for Android but left unimplemented on Windows. Without it, a device the OS already knows (bonded, or connected to another central such as a phone) cannot be reached on Windows unless it happens to be advertising. This brings WinRT to parity with Android: get a Peripheral by address, then call connect(), which already uses FromBluetoothAddressAsync.

Tested on hardware: connects to a bonded device that is connected to another central and is not advertising, then reads and writes its GATT.

### DIFF
--- a/src/winrtble/adapter.rs
+++ b/src/winrtble/adapter.rs
@@ -122,10 +122,15 @@ impl Central for Adapter {
         self.manager.peripheral(id).ok_or(Error::DeviceNotFound)
     }
 
-    async fn add_peripheral(&self, _address: &PeripheralId) -> Result<Peripheral> {
-        Err(Error::NotSupported(
-            "Can't add a Peripheral from a BDAddr".to_string(),
-        ))
+    async fn add_peripheral(&self, id: &PeripheralId) -> Result<Peripheral> {
+        if let Some(peripheral) = self.manager.peripheral(id) {
+            return Ok(peripheral);
+        }
+        // Create a peripheral straight from its address so a device the OS already knows (bonded or
+        // connected to another central) can be reached without waiting for an advertisement.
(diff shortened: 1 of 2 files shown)


tanarchytan:
The failing linux check is not from this PR.

This branch only changes `src/winrtble/adapter.rs` and `src/winrtble/peripheral.rs`. The linux job fails in the clippy step (`cargo clippy --all-features`), and the one build-breaking error is in the bluez backend, which this PR does not touch:

```
error: this loop never actually loops
  --> src/bluez/peripheral.rs:144
  note: `#[deny(clippy::never_loop)]` on by default
```

That loop returns on its first iteration, so clippy's `never_loop` lint (deny by default) rejects it. It only shows up on linux because `src/bluez` is the Linux backend and the clippy step runs on every OS with `--all-features`. macOS (corebluetooth), android, and format all pass. windows is marked cancelled only because the matrix is fail-fast and linux failed first. There are no clippy errors in the winrtble files this PR changes.

This is pre-existing under newer stable clippy (1.96, edition 2024): the most recent Rust run on master is already failing the same way. I kept the bluez code untouched to keep the PR scoped to winrtble. Glad to rerun once it is sorted, or to send a separate one-line fix for `src/bluez/peripheral.rs:144` (clippy gives a machine-applicable suggestion) if that would help.

qdot (approved):

qdot:
Rebased onto dev, will come in with 0.13.0. Thanks!

PR #466, opened 2026-07-16
/corebluetooth runtime mtu

chmorgan:
I'd like to be able to size payloads to the mtu and while the MTU branch added support the value was always 23 as apparently corebluetooth isn't updating the mtu property until after service discovery.


chmorgan:
Anything I should look at changing here?

qdot:
Nope, you're good. I brought the change in on dev after rebasing, will be in with 0.13.0

